PR #4, opened 2016-08-17
Swift3

JackieQi:
@yarneo 
Demo and Source compiled successfully. Test still has compiler errors.
Not sure if the changes make sense but hopefully can be somewhat helpful.


yarneo:
Thanks @JackieQi for the PR, this is very helpful! I will take some parts of what you did and probably change some of the behaviors like turning all the [String: AnyObject]s to [String: Any] rather than force casting value types to AnyObject to eventually become their Objective-C counterparts. Will keep you updated on a release for beta 6 hopefully tomorrow. Thanks again!

yarneo:
Updated to beta 6 + other improvements!

PR #6, opened 2016-08-19
support for beta 6 and improvements to properties typing checks

yarneo:


Mixpanel/Decide.swift
```diff
@@ -30,20 +30,20 @@ class Decide {
...
                 guard let result = decideResult else {
                     semaphore.signal()
-                    completion(response: nil)
+                    completion(nil)
                     return
                 }
 
                 var parsedNotifications = [InAppNotification]()
-                if let rawNotifications = result["notifications"] as? [[String: AnyObject]] {
+                if let rawNotifications = result["notifications"] as? [[String: Any]] {
```
yinfeiru: why `Any`? what's the edge case you are trying to handle here?

Mixpanel/DecideRequest.swift
```diff
@@ -40,10 +40,10 @@ class DecideRequest: Network {
 
     func sendRequest(distinctId: String,
                      token: String,
-                     completion: (DecideResult?) -> Void) {
+                     completion: @escaping (DecideResult?) -> Void) {
```
yinfeiru: why make it `escaping`?

Mixpanel/Mixpanel.swift
```diff
@@ -30,8 +30,8 @@ public class Mixpanel {
      You can always get the instance by calling getInstance(name)
      */
     @discardableResult
-    public class func initialize(token apiToken: String,
-                                                    launchOptions: [NSObject: AnyObject]? = nil,
+    open class func initialize(token apiToken: String,
```
yinfeiru: why do we want to make this subclass-able?

Mixpanel/Mixpanel.swift
```diff
@@ -126,6 +132,9 @@ class MixpanelManager {
     }
 
     func removeInstance(name instanceName: String) {
+        if instances[instanceName] === mainInstance {
```
yinfeiru: do we have something similar to `mainInstance` in Objective-c/Android lib? What is it for?

Mixpanel/AutomaticProperties.swift
```diff
@@ -69,8 +69,11 @@ class AutomaticProperties {
     class func deviceModel() -> String {
         var systemInfo = utsname()
         uname(&systemInfo)
-        let modelCode = withUnsafeMutablePointer(&systemInfo.machine) {
-            ptr in String(cString: UnsafePointer<CChar>(ptr))
+        let size = MemoryLayout<CChar>.size
+        let modelCode = withUnsafePointer(to: &systemInfo.machine) {
+            $0.withMemoryRebound(to: CChar.self, capacity: size) {
+                String(cString:  UnsafePointer<CChar>($0))
```
yinfeiru: extra space after `cString:`

Mixpanel/Decide.swift
yarneo: Any allows Swift value types to be sent and not only object types. If it is AnyObject, it should work as well because the object is translated to an Objective-C object, but I would rather have it fully Swift :)

Mixpanel/DecideRequest.swift
yarneo: The @escaping means that the completion closure could still run after the function has ended. Because the decide request is sent asynchronously to the server, the completion will escape the scope of the method

Mixpanel/Mixpanel.swift
yarneo: In the spirit of this being a public SDK, I wanted to leave everything open, but I'm happy to change it back to public if you think differently

Mixpanel/Mixpanel.swift
yarneo: This is a new thing that was added that allows users to get the Mixpanel instance anywhere in the code without keeping a reference to it when it was initialized

Mixpanel/AutomaticProperties.swift
yarneo: thanks :)

PR #8, opened 2016-09-01
Adding autotrack support information

kylejm:
I found by testing that Autotrack does not work with this framework, you must use the Objective-C one.


yarneo:
Thanks @kylejm! The feature we currently have for setting up tracking in the visual editor (which exists in the Objective-C SDK) is called Codeless rather than Autotrack. The difference between the two is that the web Autotrack tracks all events retroactively so even if you forget to track something, you will be able to track it retroactively in the future. The Codeless for mobile helps you track events visually but doesn't retroactively save other events you decided not to track.
If you want to rename Autotrack to Codeless I would be happy to merge the PR

yarneo:
Thanks @kylejm I appreciate the contribution! Merging the PR

PR #14, opened 2016-09-14
removePushDeviceToken fix

allu22:
Same fix as in mixpanel/mixpanel-iphone#571


yarneo:
Thanks for the quick PR! Merging it now, thanks!

PR #32, opened 2016-10-24
a/b testing feature minus docs

yarneo:


Mixpanel/Decide.swift
```diff
@@ -88,6 +93,27 @@ class Decide {
...
+                } else {
+                    Logger.debug(message: "variants check response format error")
+                }
+
+                let runningVariants = Set(self.ABTestingInstance.variants.filter { return $0.running })
+                let finishedVariants = Set(self.ABTestingInstance.variants.filter { return $0.finished })
+                decideResponse.toFinishVariants = runningVariants.subtracting(parsedVariants)
+                let newVariants = parsedVariants.subtracting(runningVariants)
+                decideResponse.newVariants = newVariants
+                let restartVariants = parsedVariants.intersection(runningVariants).intersection(finishedVariants)
```
pchien: how do we get in the state where a variant is both running and finished?

Mixpanel/VariantAction.swift
```diff
@@ -0,0 +1,395 @@
...
+                    retValue = unmanagedObject?.takeUnretainedValue()
+                }
+            }
+            targetRetValuePairs.append((object, retValue))
+        }
+        return targetRetValuePairs
+    }
+
+    class func extractAndRunMethodFromSelector(selector: Selector, implementation: IMP?, object: AnyObject, args: [Any]) -> AnyObject? {
+        if selector.description == "setImage:forState:" {
```
pchien: is there no better way to do this in swift? seems unfortunate that supporting extra changes requires sdk updates

PR #37, opened 2016-10-28
use layoutSubviews instead of didMoveToWindow for table cell ab tests…

pchien:
… so they will re-update when cells are reused

as similar to https://github.com/mixpanel/mixpanel-iphone/pull/598 as i could get

### DIFF
--- a/Mixpanel/ObjectSelector.swift
+++ b/Mixpanel/ObjectSelector.swift
@@ -106,6 +106,27 @@ class ObjectSelector: NSObject {
         return filter
     }
 
+    func selectedClass() -> AnyClass? {
+        let filter = filters.last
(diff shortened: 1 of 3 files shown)


patedit:
LGTM

Mixpanel/ObjectSelector.swift
```diff
@@ -106,6 +106,27 @@ class ObjectSelector: NSObject {
         return filter
     }
 
+    func selectedClass() -> AnyClass? {
+        let filter = filters.last
+        if let filterName = filter?.name {
```
yarneo: you can do if let filterName = filters.last?.name :)

yarneo:
LGTM

PR #46, opened 2016-12-06
Fixed crash where in the async completion block, range is out of boun…

sammarshall61:
…ds for shadowArray

### DIFF
--- a/Mixpanel.xcodeproj/project.xcworkspace/contents.xcworkspacedata
+++ b/Mixpanel.xcodeproj/project.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
--- a/Mixpanel/Flush.swift
+++ b/Mixpanel/Flush.swift
@@ -102,7 +102,11 @@ class Flush: AppLifecycle {
                                                 self.delegate?.updateNetworkActivityIndicator(false)
                                             #endif
                                             if success {
-                                                shadowQueue.removeSubrange(range)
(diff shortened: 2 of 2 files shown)


yarneo:
Thanks @sammarshall61 !! will have a look this week :)

yarneo:
hey @sammarshall61 so after reviewing the PR, it looks fine but seems a bit defensive. I have a hard time seeing why the queue would have less items and throw an out of range as the whole flush process is under a serial queue so it won't remove items whilst one flush is running. Furthermore, the asynchronous call is actually synchronous because it is used with a Semaphore. If you could provide me with more info like crash log or if you call flush explicitly that could be helpful

sammarshall61:
Hey here is a crash log, we were experiencing it for a relatively high percentage of users (our second highest crash). After deploying this fix it went away.

All the best!

[sh.whisper.whisperapp-urban_issue_3907_crash_b60160cdd739489790e7ecfa902f2d8a_2704255faae64a8eaec653e14376162a_0_v1.txt](https://github.com/mixpanel/mixpanel-swift/files/643192/sh.whisper.whisperapp-urban_issue_3907_crash_b60160cdd739489790e7ecfa902f2d8a_2704255faae64a8eaec653e14376162a_0_v1.txt)

clintron:
This might have something to do with the fact that we have two `MixpanelInstance` objects running.

yarneo:
Thanks a lot guys, I'm going forward and merging this and will at the same time try to look into the logs and reproducing with multiple instances running. Thanks again!

yarneo:
hey @sammarshall61 just one last question, did you guys edit the SDK on your end by any chance? It's just that the line numbers pointing to our code are kinda off. As an example flushQueueInBatches is only called internally from flushQueue in Flush.swift, but looks like in the crash log it is called directly from flush() in MixpanelInstance.swift? I could be totally off with my reading of the log so wanted to ask. thanks :)

PR #47, opened 2016-12-12
Support having a different server url for each MixpanelInstance

clintron:
As master stands right now, all `MixpanelInstance` objects will use the last `serverURL` that was set on any of them.

This PR initializes `MixpanelInstance` objects with a `name` property.

* `MixpanelManager` will initialize them with the `instanceName` argument.
* The `MixpanelInstance` initializer has a default value for `name` of `UUID.uuidString()` so as not to break existing code that may bypass `MixpanelManager`.

### DIFF
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -201,7 +201,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
     let trackInstance: Track
     let decideInstance: Decide
 
-    init(apiToken: String?, launchOptions: [UIApplicationLaunchOptionsKey : Any]?, flushInterval: Double, name: String=UUID().uuidString) {
+    init(apiToken: String?, launchOptions: [UIApplicationLaunchOptionsKey : Any]?, flushInterval: Double, name: String) {
         if let apiToken = apiToken, !apiToken.isEmpty {
             self.apiToken = apiToken
         }
@@ -608,7 +608,7 @@ extension MixpanelInstance {
         let defaultsKey = "trackedKey"
         if !UserDefaults.standard.bool(forKey: defaultsKey) {
             serialQueue.async() {
-                Network.trackIntegration(apiToken: self.apiToken, serverURL: self.serverURL) {
+                Network.trackIntegration(apiToken: self.apiToken, serverURL: BasePath.DefaultMixpanelAPI) {
                     (success) in
                     if success {
                         UserDefaults.standard.set(true, forKey: defaultsKey)
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -85,12 +85,9 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
     /// The base URL used for Mixpanel API requests.
     /// Useful if you need to proxy Mixpanel requests. Defaults to
     /// https://api.mixpanel.com.
-    open var serverURL: String {
-        set {
-            BasePath.MixpanelAPI = newValue
-        }
-        get {
-            return BasePath.MixpanelAPI
+    open var serverURL = BasePath.DefaultMixpanelAPI {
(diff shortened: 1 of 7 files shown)


Mixpanel/MixpanelInstance.swift
```diff
@@ -197,15 +197,20 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
...
     var serialQueue: DispatchQueue!
     var taskId = UIBackgroundTaskInvalid
-    let flushInstance = Flush()
+    let flushInstance: Flush
     let trackInstance: Track
-    let decideInstance = Decide()
+    let decideInstance: Decide
 
-    init(apiToken: String?, launchOptions: [UIApplicationLaunchOptionsKey : Any]?, flushInterval: Double) {
+    init(apiToken: String?, launchOptions: [UIApplicationLaunchOptionsKey : Any]?, flushInterval: Double, name: String=UUID().uuidString) {
```
yarneo: we don't need a default value here for name because it's already getting one in MixpanelManager

Mixpanel/MixpanelInstance.swift
```diff
@@ -85,12 +85,9 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
     /// The base URL used for Mixpanel API requests.
     /// Useful if you need to proxy Mixpanel requests. Defaults to
     /// https://api.mixpanel.com.
-    open var serverURL: String {
-        set {
-            BasePath.MixpanelAPI = newValue
-        }
-        get {
-            return BasePath.MixpanelAPI
+    open var serverURL = BasePath.DefaultMixpanelAPI {
```
yarneo: does initially setting serverURL to  BasePath.DefaultMixpanelAPI actually do something, or is this a noop?

Mixpanel/MixpanelInstance.swift
```diff
@@ -603,7 +608,7 @@ extension MixpanelInstance {
         let defaultsKey = "trackedKey"
         if !UserDefaults.standard.bool(forKey: defaultsKey) {
             serialQueue.async() {
-                Network.trackIntegration(apiToken: self.apiToken) {
+                Network.trackIntegration(apiToken: self.apiToken, serverURL: self.serverURL) {
```
yarneo: integration tracking should always be sent to the DefaultMixpanelAPI

Mixpanel/MixpanelInstance.swift
yarneo: maybe we should set the default value for that key as the defaultMixpanelAPI instead?

clintron: Right, initially setting the variable does not call `didSet`.  From apple docs:

> willSet and didSet observers are not called when a property is first initialized. They are only called when the property’s value is set outside of an initialization context.

Instead, I actually have `getServerURL()` (the function which uses `BasePath.namedBasePaths`) fall back to `BasePath. DefaultMixpanelAPI` if we don't have a `serverURL` set.

Would you like me to leave a comment?

Mixpanel/MixpanelInstance.swift
clintron: Ah good point

Mixpanel/MixpanelInstance.swift
clintron: Even though the API token may be invalid for Mixpanel?

clintron: I tried making that change, but it broke tests, so I brought this back.

Mixpanel/MixpanelInstance.swift
yarneo: no worries it's fine! I'll merge your changes tomorrow and I'll add any comments if it's unclear. Thanks again for contributing

Mixpanel/MixpanelInstance.swift
yarneo: I'll have a look after the merge, thanks!

PR #49, opened 2016-12-15
Ported identify safety check from js lib

jaredmixpanel:
Added safety check so that identify only changes the distinct_id if the given distinct_id is different from both the existing distinct_id and the alias. Required the creation of an alias property.

For reference: https://github.com/mixpanel/mixpanel-js/blob/master/src/mixpanel-core.js#L1339-L1343

### DIFF
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -42,7 +42,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
     open var distinctId = ""
 
     /// alias string that uniquely identifies the current user.
-    open var alias = ""
+    open var alias: String? = nil
 
     /// Accessor to the Mixpanel People API object.
     open var people: People!
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -41,7 +41,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
     /// distinctId string that uniquely identifies the current user.
     open var distinctId = ""
     
-    /// distinctId string that uniquely identifies the current user.
+    /// alias string that uniquely identifies the current user.
     open var alias = ""
 
     /// Accessor to the Mixpanel People API object.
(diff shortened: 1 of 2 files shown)


Mixpanel/Persistence.swift
```diff
@@ -159,14 +164,16 @@ class Persistence {
         return data as? Queue ?? []
     }
 
-    class private func unarchiveProperties(token: String) -> (InternalProperties, InternalProperties, String, String?, Queue, Set<Int>) {
+    class private func unarchiveProperties(token: String) -> (InternalProperties, InternalProperties, String, String?, String?, Queue, Set<Int>) {
```
jaredmixpanel: Ouch... not sure what the appropriate style should be here to keep this line at 140 or less.

yarneo: you do a new line after every comma in the returned values :)

Mixpanel/MixpanelInstance.swift
```diff
@@ -514,11 +522,20 @@ extension MixpanelInstance {
...
+            flush()
+        } else {
+            Logger.error(message: "alias matches distinctId - skipping api call.")
+        }
 
-        let properties = ["distinct_id": distinctId, "alias": alias]
-        track(event: "$create_alias",
-              properties: properties)
-        flush()
+
```
yarneo: please fix this :)

yarneo:
overall looks fine. I'm assuming you tested this on different scenarios?

Mixpanel/MixpanelInstance.swift
```diff
@@ -40,6 +40,9 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
 
     /// distinctId string that uniquely identifies the current user.
     open var distinctId = ""
+
```
jaredmixpanel: It looks to me like this is fixed and these violations are outdated. Maybe I'm missing something.

Mixpanel/MixpanelInstance.swift
jaredmixpanel: As best as I can tell this is fixed, maybe I'm missing it though.

jaredmixpanel:
@yarneo  Yes, I've done some basic testing, running through different scenarios of changing the distinct_id and the alias, ensuring that the logic behaves as intended.

PR #53, opened 2016-12-16
Update settings for Xcode 8.2

mikeyLatch:
For Xcode 8.2, the project should also have the "Use legacy Swift Language" set.


yarneo:
thanks! pushed a commit that fixes the versioning for all the targets. will send out a release today

PR #56, opened 2017-01-19
Add optional question mark

smswz:
The gesture recognizer code was missing a `?.` for non-simulator builds (Archive)


yarneo:
Thanks for finding that @smswz ! merging.

PR #58, opened 2017-02-02
In-Apps 2.0 implementation

yarneo:


Mixpanel/Info-tvOS.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.1.2</string>
+	<string>2.1.4</string>
```
patedit: Is this intentional? Where did 2.1.3 go? :P

Mixpanel/TakeoverNotification.swift
```diff
@@ -0,0 +1,85 @@
...
+            Logger.error(message: "invalid notification fade image boolean")
+            return nil
+        }
+        self.shouldFadeImage    = shouldFadeImage
+        imageURL = URL(string: imageURL.absoluteString.appendSuffixBeforeExtension(suffix: "@2x"))!
+
+    }
+}
+
+extension String {
```
patedit: this is a good candidate to go somewhere else (i.e. it's not in-app application dependent)

MixpanelDemo/MixpanelDemoTests/MixpanelNotificationTests.swift
```diff
@@ -14,93 +14,90 @@ import Nocilla
 
 class MixpanelNotificationTests: MixpanelBaseTests {
```
patedit: I think we should test for all the possible fields in an in-app (including button's field and whether `buttons` is an array, for example).
We should also include tests for mini in-apps (probably a separate function)

Mixpanel/Decide.swift
```diff
@@ -67,8 +67,14 @@ class Decide {
                 var parsedNotifications = [InAppNotification]()
                 if let rawNotifications = result["notifications"] as? [[String: Any]] {
                     for rawNotif in rawNotifications {
-                        if let notification = InAppNotification(JSONObject: rawNotif) {
-                            parsedNotifications.append(notification)
+                        if let notificationType = rawNotif["type"] as? String {
+                            if notificationType == InAppType.takeover.rawValue,
+                                let notification = TakeoverNotification(JSONObject: rawNotif) {
+                                parsedNotifications.append(notification)
```
yarneo: that's what that line does, it's called conditional chaining in swift, if let notification is null the conditional won't pass

Mixpanel/Info-tvOS.plist
yarneo: i thought we would do a release before the in-apps release so jumped it by 2 :)

Mixpanel/Decide.swift
patedit: whoah, the indenting got me confused, did not see the if

PR #62, opened 2017-02-10
MixpanelInstance updates: Notification Payload access

rresma2:
Hey there,

My name is Rob Resma. I'm an iOS developer for a company called Asurion. We currently have a partnership with Mixpanel and have been working closely with two of your colleagues Emmett Faricy and Bhanu Shandilya to come up with a solution for a problem we're currently tackling in our iOS app. Currently, the Mixpanel swift library doesn't offer any customization of the InAppNotification UI and presentation logic. Emmett suggested that I fork your library and explore different solutions from within your library.

This method that I have created is to get access to the InApp notification payload that comes from your server. I noticed that much of the InAppNotification code is pretty coupled, so I couldn't figure out a way to build on your library without touching a large portion of the related classes. I figured this change would allows us to achieve our end goal while touching as little of your library as possible.

Take a look and let me know what you think and if this PR needs anything else.

Best,

Rob


rresma2:
@yarneo I was wondering if you could see this. I'm not sure how to add you or anyone else as a reviewer.

yarneo:
Thanks @rresma2 I can see this :) I'll be able to take a look at it on Tuesday. Hope that's ok!

rresma2:
Ok sweet! 👍  Thanks @yarneo

yarneo:
Hi @rresma2!
As you can see we now support customization of our in-apps! Have a look at the in-app builder in mixpanel.com and you will now see a tab for "custom". There you can add a secondary button, change colors, etc. Let me know if that helps you with your needs for better in-apps

rresma2:
Hi, @yarneo I did see the changes. I noticed a few of the changes do address some of the design requirements we have!

1. transparent background
2. custom colors
3. secondary 

There are however, other UI customizations that our team wants to do. A few of them are:

1. Having the title and image interchangeable
2. Custom fonts for the text
3. No (x) dismiss button at the top right corner
4. White background encapsulating the content w/ corner radii 

I'm not sure how much bandwidth you guys have to implement these customizations, which is why I thought the best way to handle our specifications by just getting a hold of the payload and handling the display ourselves. If you want to look more into these UI customizations, I'm also more than happy to help.

yarneo:
That makes sense! I'm curious to know how you make use of the payload. Do you basically fetch it and add extra fields then display the in-app using your own customized XIB file?

rresma2:
Yep, exactly. For my points 2, 3, and 4, we use interface builder to set the corner radii, fonts, and (x) button. As far as for interchanging the title and image, we're currently working around complete control by using Mixpanel tweaks to AB Test the two layouts. Also, instead of directly calling UIApplication.shared.open(url), we're actually passing a custom scheme that allows us to navigate to a certain page within our app after taking the action.

yarneo:
Thanks @rresma2 ! Merging!

rresma2:
Awesome. Thank you @yarneo :)

rresma2:
@yarneo I'm not familiar with how to create and update CocoaPods. What does it take to update the Pod version with the latest code?

yarneo:
We would have to send out a new release on our end but we won't be doing that until we have some more improvements to send out with that release. What you could do in the meanwhile is ask the pod to point to the master branch like so:
`pod 'Mixpanel-swift', :git => 'https://github.com/mixpanel/mixpanel-swift.git'`

rresma2:
Sure that makes sense. We'll do that for now, thanks!

rresma2:
Sure that makes sense. We'll do this for now, thanks!

PR #66, opened 2017-03-01
Added App Extension and Mac OS support!

yarneo:


Mixpanel-swift-appex.podspec
```diff
@@ -0,0 +1,21 @@
...
+  s.name = 'Mixpanel-swift-appex'
+  s.version = '2.1.3'
+  s.module_name = 'MixpanelAppex' 
+  s.license = 'Apache License, Version 2.0'
+  s.summary = 'Mixpanel tracking library for iOS (Swift) App Extensions'
+  s.homepage = 'https://mixpanel.com'
+  s.author       = { 'Mixpanel, Inc' => 'support@mixpanel.com' }
+  s.source       = { :git => 'https://github.com/mixpanel/mixpanel-swift.git',
+                     :tag => "v#{s.version}",
+                     :branch => "appextension-support" }
```
pchien: should this branch be removed?

Mixpanel/AutomaticProperties.swift
```diff
@@ -47,7 +65,9 @@ class AutomaticProperties {
             p["$ios_app_release"] = infoDict["CFBundleShortVersionString"]
         }
         p["$ios_device_model"]  = AutomaticProperties.deviceModel()
+        #if !MAC_OS
         p["$ios_version"]       = UIDevice.current.systemVersion
```
pchien: should we put the macOS version in this too? since we're putting other random properties in $ios_blah

Mixpanel/MixpanelInstance.swift
```diff
@@ -235,9 +246,31 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate {
...
     }
+    #else
+    init(apiToken: String?, flushInterval: Double, name: String) {
+        if let apiToken = apiToken, !apiToken.isEmpty {
+            self.apiToken = apiToken
+        }
 
+        self.name = name
+
+        flushInstance = Flush(basePathIdentifier: name)
```
pchien: can we do the common stuff in another function? lots of things duplicated here and above

Mixpanel/Persistence.swift
```diff
@@ -195,7 +242,37 @@ class Persistence {
                 peopleUnidentifiedQueue,
                 shownNotifications)
     }
+    #else
+    class private func unarchiveProperties(token: String) -> (InternalProperties,
```
pchien: same idea with the unarchive/archive stuff wrt duplicate code

Mixpanel-swift-appex.podspec
yarneo: yep!

Mixpanel/AutomaticProperties.swift
yarneo: sure!

Mixpanel/MixpanelInstance.swift
yarneo: if i do so, all the params that are in an intermediate method would have to become not consts but rather vars (let vs var) and i would hate to remove the immutability for them

Mixpanel/Persistence.swift
yarneo: sure

Mixpanel/MixpanelInstance.swift
pchien: i don't get it but i don't understand swift that well so probably fine then

pchien:
new changes lgtm

PR #77, opened 2017-03-28
Clear up `pod` usage

mxcl:
If you want Mixpanel in both your app *and* your app-extension, you need both pods specified in the correct targets in your `Podfile`.


yarneo:
Thanks for the edit, LGTM!

PR #79, opened 2017-04-08
Fix compile errors when DECIDE is not unset

brow:
Without this change, if you remove `-D DECIDE` from `OTHER_SWIFT_FLAGS` in the build settings, then compilation fails. I hope this is helpful!


yarneo:
This is great, thanks @brow ! Merging.

PR #84, opened 2017-05-10
Cast campaign_id and message_id to ints

wgins:
`mpPayload["m"]` and `mpPayload["c"]` are ints. To avoid indavertently evaluating `c` and `m` to `nil` we cast to Int instead. This is consistent with our [Obj-C SDK](https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/Mixpanel.m#L412-L414).


yarneo:
LTST

PR #86, opened 2017-05-13
Update unit tests for push notification open tracking

wgins:
Updating unit tests to reflect ints in remote notification payload for campaign_id and message_id keys.
cc: @yarneo


yarneo:
cheers

PR #91, opened 2017-06-12
Allow tracking journey_id in push open 

Nicomoto:
#### Push payload
```
{
	‘mp’: {
                 ‘c’: $campaign_id,
 	         ‘m’: $message_id,
        }
}
```
 
#### Track call
```
{
	event: $campaign_received,
  	properties: {
		‘campaign_id': $campaign_id,
		‘message_id’: $message_id
	}
}
```
 
#### Push payload
```
{
	‘mp’: {
                ‘c’: $campaign_id,
 	        ‘m’: $message_id,
		‘additional_params_1’: $param1,
	        ‘additional_params_2’: $param2,
        }
}
```
 
### Track call
```
{
	event: $campaign_received,
  	properties: {
		‘campaign_id': $campaign_id,
		‘message_id’: $message_id,
		‘additional_params_1’: $param1,
		‘additional_params_2’: $param2,
	}
}
```

### DIFF
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -896,6 +896,11 @@ extension MixpanelInstance {
         if let mpPayload = userInfo["mp"] as? InternalProperties {
             if let m = mpPayload["m"], let c = mpPayload["c"] {
                 var properties = Properties()
+                for (key, value) in mpPayload {
+                    if key != "m" && key != "c" {
+                        properties[key] = value
+                    }
+                }
                 properties["campaign_id"]  = c as? Int
                 properties["message_id"]   = m as? Int
                 properties["message_type"] = "push"
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -898,9 +898,10 @@ extension MixpanelInstance {
                 var properties = Properties()
                 for (key, value) in mpPayload {
                     if key != "m" && key != "c" {
-                        if let typedValue = value as? MixpanelType {
-                            properties[key] = typedValue;
-                        }
+                        // Check Int first, since a number in the push payload is parsed as __NCSFNumber
+                        // which fails to convert to MixpanelType.
+                        if let typedValue = value as? Int { properties[key] = typedValue }
+                        if let typedValue = value as? MixpanelType { properties[key] = typedValue }
                     }
                 }
                 properties["campaign_id"]  = c as? Int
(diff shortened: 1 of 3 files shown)


yarneo:
is there a reason for adding all additional params rather than just "journey_id"?

yarneo:
This also doesn't compile, would be great if you could set up the environment and see that it compiles and tests are passing (Peter should be able to help you with that if he hasn't already)

Nicomoto:
I've tested this with the SDK, and the demo app today. I had to add NSNumber to mixpanel type since it seems like the json parsing of the push payload returns that.

The reason for adding all additional params is so we don't have to go through SDK cycles to add new params in the future, it would just be server controlled.

yarneo:
if you have added NSNumber to MixpanelType, could you please test with this implementation:
`open func trackPushNotification(_ userInfo: [AnyHashable: Any],
                                      event: String = "$campaign_received") {
        if var mpPayload = userInfo["mp"] as? Properties {
            if let m = mpPayload["m"], let c = mpPayload["c"] {
                mpPayload["campaign_id"]  = c
                mpPayload["message_id"]   = m
                mpPayload["message_type"] = "push"
                mpPayload.removeValue(forKey: "c")
                mpPayload.removeValue(forKey: "m")
                track(event: event,
                      properties: mpPayload)
            } else {
                Logger.info(message: "malformed mixpanel push payload")
            }
        }
    }
`

That way we won't have to iterate and do a check every iteration. Thanks!

Nicomoto:
I think that should work with our current values. The reason I did not prefer that approach is that its possible to break tracking completely (and perhaps following things) by passing some malformed value from the server. I was trying to model it such that one malformed value will not break tracking of all others. 

Don't feel too strongly about this, so I can make this change depending on your thoughts

yarneo:
I understand your concern! do you have any idea on what future params we will be receiving? It would be great to have a clearer picture so we can prepare better. Because Swift is strongly typed, if we receive unsupported types from the server, we won't accept it; which i think is good. In general, I honestly am not the biggest fan of adding NSNumber to the MixpanelType list because i'd prefer people to add properties to their events with native Swift types (and not Objective-C ones). But if we do add it to MixpanelType, let's make it simpler.

I think an alternative is to either iterate over the params we receive and convert them to Swift native types, or to add only journey_id currently, if in fact you don't foresee any additional params being added in the near future.

Nicomoto:
Talked offline, and decided to handle NSNumber being passed from the json parsing in this function alone instead of adding to MixpanelType as done with the other params.

yarneo:
LGTM!

Issue #87, opened 2017-05-22
Not necessary to split into two pods


mxcl:
I have examined why you split the framework into two and it isn't necessary.

It is certainly quite inconvenient, not to mention inefficient for my apps that you do. So it seems more ideal that it would not be done.

The various hooks that the `Mixpanel-swift` does that `Mixpanel-swift-appex` doesn't could all be decided by examining the runtime and not with compile-time `#ifdefs`.

Then the extension can be marked app-extension-safe and you only need one framework.

yarneo:
Hey @mxcl, agreed that it's a pain point.

When you suggest checking if it's an app extension at runtime, are you referring to see if it can perform .shared on UIApplication or did you have something else in mind?

mxcl:
This issue is actually causing a crash for me now. As a result I am going to switch to only using the appex version.

The crash is because I have a framework that links to the appex version because it is used by the app and several extensions. The main app uses the non-appex version. The normal version gets its keys set but the appex version is then used by the framework to do tracking. You have a completely unreasonable `fatalError` when `mainInstance` is `nil` and I now look bad to my client.

You should fix this. I wonder why you ever decided this was a reasonable way to do this in the first place.

yarneo:
Thanks @mxcl, I understand your frustration. The fatalError is there more importantly so people integrate Mixpanel correctly by initializing before calling any of the instance's methods/properties. This usually is caught when testing and should arise when any Mixpanel track event or even identify is called. We make the instance easy to reach anywhere in your code (without needing to pass it around) by calling `mainInstance()` but in order to use that, one has to initialize first. I do understand your confusion that if you initialized in the main app, you would want it to be initialized in the appex as well. However, you do need to initialize in both as they are separate and sandboxed from one another. As a side note, because `mainInstance()` is usually called many times, we tried to avoid having it an optional so it could return `nil` on extreme cases and then forcing people to use `mainInstance()?` on a regular basis.

I will look into a possible solution in the next two weeks on how we can combine both pods into one without causing this Cocoapods/Apple issue to arise: https://github.com/CocoaPods/CocoaPods/issues/6711
As you said, this seems like something that can currently be done only if an app extension is identifiable at runtime.

mxcl:
> The fatalError is there more importantly so people integrate Mixpanel correctly by initializing before calling any of the instance's methods/properties. This usually is caught when testing and should arise when any Mixpanel track event or even identify is called.

[...]

PR #92, opened 2017-06-13
Have app extension support as part of the main target

yarneo:
we now check if it is an app extension in runtime (and not with compilation flags) and hence allowing us to consolidate the appextension target and the iOS target into one, and also remove the need of a separate Pod to work around this issue: https://github.com/CocoaPods/CocoaPods/issues/6711
this also solves this issue:
https://github.com/mixpanel/mixpanel-swift/issues/87


Mixpanel.xcodeproj/project.pbxproj
```diff
@@ -1026,6 +939,7 @@
 		E11594861CFF1491007F8B4F /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
```
pchien: the internet suggests this removes things that aren't allowed in app extensions at compile time. does that mean that if people want to use app extensions they lose a bunch of functionality in their main app as well? or is this a setting that can be different between their app and the extension?

yarneo: AFAIK, this is set in order to not allow the compiler or linker to let developers access APIs they shouldn't access in appex (such as UIApplication.shared). In our case because we are accessing these APIs (implicitly) only in the main app and not in the appex (which we figure out during runtime), we are fine. However, i do prefer to be less strict as there may be unknowns and I've seen that this flag is automatically turned on for Appex apps anyway, so i'll set it as NO for now.

pchien: oh as long as it only affects the appex part of the app i'm okay with either setting for this

PR #98, opened 2017-06-26
Fix compilation without DECIDE flag

arthur-here:
This PR fixes compilation without DECIDE flag


yarneo:
Thanks for doing this @arthur-here !

PR #100, opened 2017-06-28
swift 4 (not to be merged)

yarneo:


Mixpanel/Variant.swift
```diff
@@ -157,7 +157,7 @@ class Variant: NSObject, NSCoding {
         if object === self {
             return true
         } else {
-            return self.ID == object.ID
+            return self.ID == object.ID && self.actions == object.actions
```
pchien: it looks like VariantAction's equal check only checks the name. is that enough to check if they're actually doing the same thing? i.e. if someone changes an action in their ab test will the name that's sent back change too?

also do we want to do this on a separate branch to fix it on master earlier than september?

matt-loflin:
Proposed fix for Xcode 9 Beta 3 error:
Add Objective-C name to line 117 of TweakPersistency.swift
```@objc(TweakDiskPersistencyData) private final class Data: NSObject, NSCoding {```

Mixpanel/Variant.swift
yarneo: it actually does change the name of the action after every change!

yarneo: and i will push out a fix earlier to master for this

matt-loflin:
@yarneo Any chance you could merge #105 into this branch?

PR #101, opened 2017-06-29
Don't require identify to merge $experiments People prop

jaredmixpanel:
Same change that was made in mixpanel-iphone lib here: https://github.com/mixpanel/mixpanel-iphone/pull/693


yarneo:
thanks @jaredmixpanel !

PR #104, opened 2017-07-06
Seperate queues

herosang:
Separated network requests from tracking requests.

Network requests are now sent to a new serial queue networkQueue.

2 new queues (copyEventQueue, and copyPeopleQueue) are now used for flushing.

Changed a bunch of the tests to accurately reflect the changes that were made.


Mixpanel/MixpanelInstance.swift
```diff
@@ -724,8 +729,8 @@ extension MixpanelInstance {
                                             peopleUnidentifiedQueue: people.unidentifiedQueue,
                                             shownNotifications: decideInstance.notificationsInstance.shownNotifications,
                                             automaticEventsEnabled: decideInstance.automaticEventsEnabled)
-        Persistence.archive(eventsQueue: eventsQueue,
-                            peopleQueue: people.peopleQueue,
+        Persistence.archive(eventsQueue: copyEventsQueue + eventsQueue,
```
yarneo: is there a reason why the copy is appended to the normal queue?

Mixpanel/MixpanelInstance.swift
```diff
@@ -845,19 +850,42 @@ extension MixpanelInstance {
...
+            
             if let shouldFlush = self.delegate?.mixpanelWillFlush(self), !shouldFlush {
                 return
             }
+            
+            self.trackingQueue.sync {
+                self.copyEventsQueue = self.eventsQueue
+                self.copyPeopleQueue = self.people.peopleQueue
+
+                self.eventsQueue.removeAll()
```
yarneo: could there be a case where eventsQueue or peopleQueue are written or read during this time?

yarneo: I know that in the Obj-C PR there was a synchronized every time we did a read/write to these queues so curious to know how we solved that issue

Mixpanel/MixpanelInstance.swift
```diff
@@ -845,19 +850,42 @@ extension MixpanelInstance {
      - parameter completion: an optional completion handler for when the flush has completed.
      */
     open func flush(completion: (() -> Void)? = nil) {
-        serialQueue.async() {
+        
+        self.archive()
+
+        networkQueue.async() {
```
yarneo: we should probably still keep the flush ordered. so we should wait for the trackingQueue to finish the tasks it set to do before the flush was invoked and only then issue the flush

Mixpanel/MixpanelInstance.swift
```diff
@@ -1092,7 +1120,7 @@ extension MixpanelInstance: InAppNotificationsDelegate {
 
     // MARK: - Decide
     func checkDecide(forceFetch: Bool = false, completion: @escaping ((_ response: DecideResponse?) -> Void)) {
-        serialQueue.async {
+        trackingQueue.async {
```
yarneo: this should be part of the network queue but also wait for the tracking queue to finish its current tasks

yarneo:
we should make sure that when we reset() we wait for the network queue to finish its tasked before resetting

Mixpanel/MixpanelInstance.swift
pchien: i agree. this was the reason why we did serial queue async -> network queue async. it makes it so that if users do track and then flush it guarantees that track is in the flush, like they expect

Mixpanel/MixpanelInstance.swift
```diff
@@ -845,19 +850,42 @@ extension MixpanelInstance {
...
-            self.flushInstance.flushEventsQueue(&self.eventsQueue,
+            self.flushInstance.flushEventsQueue(&self.copyEventsQueue,
                                                 automaticEventsEnabled: false)
             #endif
-            self.flushInstance.flushPeopleQueue(&self.people.peopleQueue)
+            self.flushInstance.flushPeopleQueue(&self.copyPeopleQueue)
+
+            self.trackingQueue.sync {
+                self.eventsQueue = self.copyEventsQueue + self.eventsQueue
+                self.people.peopleQueue = self.copyPeopleQueue + self.copyPeopleQueue
```
pchien: this should be `self.copyPeopleQueue + self.people.peopleQueue` right?

Mixpanel/MixpanelInstance.swift
```diff
@@ -224,8 +224,11 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     var apiToken = ""
     var superProperties = InternalProperties()
     var eventsQueue = Queue()
+    var copyEventsQueue = Queue();
```
pchien: can we rename these? `copyEventsQueue` and `copyPeopleQueue` don't tell me much about what they actually are since they're not exactly copies of the queues

Mixpanel/MixpanelInstance.swift
```diff
@@ -756,7 +761,7 @@ extension MixpanelInstance {
                                             peopleDistinctId: people.distinctId,
                                             peopleUnidentifiedQueue: people.unidentifiedQueue)
         Persistence.archive(eventsQueue: copyEventsQueue + eventsQueue,
```
pchien: this one also needs to be renamed to `flushEventsQueue`

Mixpanel/MixpanelInstance.swift
```diff
@@ -686,6 +686,11 @@ extension MixpanelInstance {
      */
     open func reset() {
         trackingQueue.async() {
+            
+            self.networkQueue.sync {
+                return;
```
pchien: i think you can deadlock here if you call `flush` and `reset` and both are added to the queues before either executes. then you have flush happening on the network queue and reset happening on the tracking queue, with reset only able to complete after the network queue is emptied and flush needing to do a sync operation on the tracking queue so it can't complete

Mixpanel/MixpanelInstance.swift
```diff
@@ -1120,7 +1124,10 @@ extension MixpanelInstance: InAppNotificationsDelegate {
 
     // MARK: - Decide
     func checkDecide(forceFetch: Bool = false, completion: @escaping ((_ response: DecideResponse?) -> Void)) {
-        trackingQueue.async {
+        networkQueue.async {
+            self.trackingQueue.sync {
+                return;
```
pchien: this also can have a deadlock with reset for the same reason

Mixpanel/MixpanelInstance.swift
```diff
@@ -1124,14 +1122,14 @@ extension MixpanelInstance: InAppNotificationsDelegate {
 
     // MARK: - Decide
     func checkDecide(forceFetch: Bool = false, completion: @escaping ((_ response: DecideResponse?) -> Void)) {
-        networkQueue.async {
-            self.trackingQueue.sync {
-                return;
-            }
-            self.decideInstance.checkDecide(forceFetch: forceFetch,
+        self.trackingQueue.async {
+            self.networkQueue.async {
```
pchien: i think because you flipped tracking and network in both reset and here it can still deadlock - double check this though

pchien:
also i don't see any changes for the backgrounding stuff. since we changed flush to be on network (and then on tracking, but it's async on network first) did you ensure that the background task will wait for flush to finish?

pchien:
what about other archive methods like `archiveProperties`?

Mixpanel/MixpanelInstance.swift
```diff
@@ -621,12 +621,14 @@ extension MixpanelInstance {
...
             if !self.people.unidentifiedQueue.isEmpty {
+                objc_sync_enter(self)
                 for var r in self.people.unidentifiedQueue {
                     r["$distinct_id"] = self.distinctId
                     self.people.peopleQueue.append(r)
                 }
                 self.people.unidentifiedQueue.removeAll()
-                Persistence.archivePeople(self.flushPeopleQueue + self.people.peopleQueue, token: self.apiToken)
+                objc_sync_exit(self)
+                Persistence.archivePeople(self.people.flushPeopleQueue + self.people.peopleQueue, token: self.apiToken)
```
pchien: @yarneo you said it copies `self.people.peopleQueue` here right? it seems unlikely that's atomic but do you know off the top of your head? if it isn't i think the copying needs to be inside `objc_sync` as well

yarneo: swift is non-atomic so we will need any copy (set) be under a thread-safe encapsulation

Mixpanel/MixpanelInstance.swift
```diff
@@ -232,7 +232,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     var taskId = UIBackgroundTaskInvalid
     #endif // os(OSX)
     let flushInstance: Flush
-    let trackInstance: Track
+    var trackInstance: Track! = nil
```
pchien: i thought the `!` in swift meant it couldn't be `nil`? am i wrong?

yarneo: correct, and it's bad practice to have ! in general

yarneo: (it can be nil only it will be explicitly unwrapped == unsafe)

Mixpanel/MixpanelInstance.swift
```diff
@@ -1135,7 +1140,7 @@ extension MixpanelInstance: InAppNotificationsDelegate {
     // MARK: - Decide
     func checkDecide(forceFetch: Bool = false, completion: @escaping ((_ response: DecideResponse?) -> Void)) {
         self.trackingQueue.async {
-            self.networkQueue.async {
+            self.networkQueue.sync {
```
pchien: does this mean decide blocks the tracking queue until the network request completes?

Mixpanel/MixpanelInstance.swift
```diff
@@ -248,20 +249,20 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
         #if DECIDE
             decideInstance = Decide(basePathIdentifier: name)
         #endif // DECIDE
-        trackInstance = Track(apiToken: self.apiToken, mixpanelInstance: self)
+        self.readWriteLock = ReadWriteLock(label: "readWrite")
```
pchien: why do we need to pass the label in?

Mixpanel/ReadWriteLock.swift
```diff
@@ -0,0 +1,31 @@
...
+    }
+    
+    func read(closure: () -> ()) {
+        self.conccurentQueue.sync {
+            closure();
+        }
+    }
+    
+    func write(closure: () -> ()) {
+        self.conccurentQueue.sync(flags: .barrier, execute: {
```
pchien: this is probably fine. i don't know why apple didn't document `.barrier` https://developer.apple.com/documentation/dispatch/dispatchworkitemflags/1780674-barrier?changes=latest_minor but i assume it works the same as https://developer.apple.com/documentation/dispatch/1452917-dispatch_barrier_sync

Mixpanel/ReadWriteLock.swift
```diff
@@ -0,0 +1,31 @@
...
+//  ReadWriteLock.swift
+//  Mixpanel
+//
+//  Created by Hairuo Sang on 8/9/17.
+//  Copyright © 2017 Mixpanel. All rights reserved.
+//
+
+import Foundation
+
+class ReadWriteLock{
```
pchien: can you create a unit test for this? just to guarantee that it works the way we want

pchien:
i think it's all good now. the latest changes look fine anyway

Mixpanel/Flush.swift
```diff
@@ -141,6 +141,7 @@ class Flush: AppLifecycle {
                                             semaphore.signal()
                 })
                 _ = semaphore.wait(timeout: DispatchTime.distantFuture)
+
```
yarneo: let's remove this line addition

Mixpanel/InAppNotifications.swift
```diff
@@ -20,15 +20,20 @@ enum InAppType: String {
 }
 
 class InAppNotifications: NotificationViewControllerDelegate {
-
+    
+    let lock:ReadWriteLock
```
yarneo: space after the `:` please !

Mixpanel/InAppNotifications.swift
```diff
@@ -54,10 +59,12 @@ class InAppNotifications: NotificationViewControllerDelegate {
     }
 
     func markNotificationShown(notification: InAppNotification) {
-        Logger.info(message: "marking notification as seen: \(notification.ID)")
+        self.lock.write {
```
yarneo: `self` is not needed other than in `init` or in a block

Mixpanel/ReadWriteLock.swift
```diff
@@ -0,0 +1,30 @@
...
```
yarneo: space between class name and `{`

Mixpanel/ReadWriteLock.swift
```diff
@@ -0,0 +1,30 @@
...
+    let conccurentQueue: DispatchQueue
```
yarneo: i think it's concurrent ?

Mixpanel/Track.swift
```diff
@@ -52,48 +56,64 @@ class Track {
...
+                eventsQueue.remove(at: 0)
+            }
         }
+
     }
 
     func registerSuperProperties(_ properties: Properties, superProperties: inout InternalProperties) {
-        assertPropertyTypes(properties)
-        superProperties += properties
+        self.lock.write {
```
yarneo: make sure we aren't calling `self` when not needed

MixpanelDemo/MixpanelDemoTests/MixpanelDemoTests.swift
```diff
@@ -185,27 +190,27 @@ class MixpanelDemoTests: MixpanelBaseTests {
                            kTestToken, "incorrect project token in people record")
             XCTAssertEqual(mixpanel.people.peopleQueue.last?["$distinct_id"] as? String,
                            distinctId, "distinct id not set properly on unidentified people record")
-            var p: InternalProperties = mixpanel.people.peopleQueue.last?["$set"] as! InternalProperties
-            XCTAssertEqual(p["p1"] as? String, "a", "custom people property not queued")
-            assertDefaultPeopleProperties(p)
+           // var p: InternalProperties = mixpanel.people.peopleQueue.last?["$set"] as! InternalProperties
```
yarneo: why did we comment these? if we are removing them let's remove them

Mixpanel/MixpanelInstance.swift
```diff
@@ -224,14 +224,17 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     var apiToken = ""
     var superProperties = InternalProperties()
     var eventsQueue = Queue()
+    var flushEventsQueue = Queue();
```
yarneo: remove `;`

Mixpanel/MixpanelInstance.swift
```diff
@@ -242,22 +245,24 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
             self.apiToken = apiToken
         }
         self.name = name
+        self.readWriteLock = ReadWriteLock(label: "globalLock")
         flushInstance = Flush(basePathIdentifier: name)
         #if DECIDE
-            decideInstance = Decide(basePathIdentifier: name)
+            decideInstance = Decide(basePathIdentifier: name, lock:self.readWriteLock)
```
yarneo: space after `:`

Mixpanel/Persistence.swift
```diff
@@ -77,14 +77,19 @@ class Persistence {
     #endif // DECIDE
 
     class func archiveEvents(_ eventsQueue: Queue, token: String) {
+        objc_sync_enter(self)
```
yarneo: if we are using a lock now, do we still need these to use a different lock on self? do we not want them synced with the other stuff?

PR #105, opened 2017-07-11
Fix build error in swift4wip branch (Xcode 9 beta 3)

matt-loflin:
Use the @objc(name) attribute to change the name of the private nested Data class as it’s exposed to Objective-C code.


yarneo:
thanks for committing this @matt-loflin !

PR #106, opened 2017-07-13
swift fixes

yarneo:
this fixes the crash we have been seeing with the UITabSwappableImageView.
this also fixes the modification of existing a/b tests.


yarneo:
Just tested it with Obj-C, seems like it's an issue there as well. I'll open a PR for it.

PR #107, opened 2017-07-14
persist identity in userdefaults. also xcode9 warnings fixed.

yarneo:


Mixpanel/Persistence.swift
```diff
@@ -118,8 +118,21 @@ class Persistence {
...
+            return
         }
 
+        addSkipBackupAttributeToItem(at: path)
+    }
+
+    class private func addSkipBackupAttributeToItem(at path: String) {
+        var url = URL.init(fileURLWithPath: path)
+        var resourceValues = URLResourceValues()
+        resourceValues.isExcludedFromBackup = true
```
pchien: what's the reason behind skipping backups? feels like backups might be nice to transfer things like super props to new phones. probably don't want backups for queue archives though

yarneo: sorry, forgot to add context: https://github.com/mixpanel/mixpanel-iphone/issues/565

Issue #111, opened 2017-07-24
Crash when using with Khan/SwiftTweaks


tomekh7:
There is a conflict in naming while unarchiving in `loadFromDisk` function which causes a crash. It looks like you copied Khan/SwiftTweaks and modified it without changing location of the file where tweaks data is stored. So when we use Khan/SwiftTweaks as well, we have crash even if we don't enable Khan's tweaks for production build, but we use Mixpanel experiments for A/B testing.

> The application crash with the message: `Terminating app due to uncaught exception 'NSInvalidUnarchiveOperationException', reason: '*** -[NSKeyedUnarchiver decodeObjectForKey:]: cannot decode object of class (Data) for key (root); the class may be defined in source code or a library that is not linked'

yarneo:
Thanks for reporting this @tomekh7 and especially for finding out the root cause! I will be working on this and sending out a new release by tomorrow. 

Thanks again and sorry for the inconvenience. Will update this thread accordingly on progress.

yarneo:
fixed in v2.2.0!

PR #112, opened 2017-07-24
change db location of tweaks

yarneo:
change db location of tweaks, and execute tweaks from cached mixpanel archives after Tweaks have been initialized


Mixpanel/TweakPersistency.swift
```diff
@@ -64,7 +64,7 @@ private final class TweakDiskPersistency {
 
 	private static func fileURLForIdentifier(_ identifier: String) -> URL {
         return try! FileManager().url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
-			.appendingPathComponent("SwiftTweaks")
```
tomekh7: That should resolve issue #111 Thanks

PR #125, opened 2017-08-21
properties for $campaign_open for the cta url and if the button was p…

pchien:
…rimary or secondary. also now always track $campaign_open if they tap on either button or mini, instead of only if there was a cta url

### DIFF
--- a/Mixpanel/InAppNotifications.swift
+++ b/Mixpanel/InAppNotifications.swift
@@ -11,7 +11,7 @@ import UIKit
 
 protocol InAppNotificationsDelegate {
     func notificationDidShow(_ notification: InAppNotification)
-    func notificationDidCTA(_ notification: InAppNotification, event: String)
+    func trackNotification(_ notification: InAppNotification, event: String, properties: Properties?)
 }
 
 enum InAppType: String {
@@ -66,7 +66,10 @@ class InAppNotifications: NotificationViewControllerDelegate {
         miniNotificationVC.show(animated: true)
 
         DispatchQueue.main.asyncAfter(deadline: .now() + miniNotificationPresentationTime) {
-            self.notificationShouldDismiss(controller: miniNotificationVC, callToActionURL: nil)
+            self.notificationShouldDismiss(controller: miniNotificationVC,
+                                           callToActionURL: nil,
+                                           shouldTrack: false,
+                                           additionalTrackingProperties: nil)
         }
         return true
     }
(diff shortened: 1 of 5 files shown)


Mixpanel/InAppNotifications.swift
```diff
@@ -79,20 +82,32 @@ class InAppNotifications: NotificationViewControllerDelegate {
...
+                                   additionalTrackingProperties: Properties?) -> Bool {
         if currentlyShowingNotification?.ID != controller.notification.ID {
             return false
         }
 
         let completionBlock = {
+            if shouldTrack {
+                var properties = additionalTrackingProperties
+                if let urlString = callToActionURL?.absoluteString {
+                    if properties == nil {
```
yarneo: is there a reason why we don't leave this nil and send it like that? there is a unwrapping in trackNotifcation

Mixpanel/InAppNotifications.swift
```diff
@@ -79,20 +82,32 @@ class InAppNotifications: NotificationViewControllerDelegate {
...
+                        properties = [:]
+                    }
+                    properties!["url"] = urlString
```
yarneo: `if let properties = properties { properties["url"] = urlString }`

Mixpanel/InAppNotifications.swift
```diff
@@ -79,20 +82,32 @@ class InAppNotifications: NotificationViewControllerDelegate {
     }
 
     @discardableResult
-    func notificationShouldDismiss(controller: BaseNotificationViewController, callToActionURL: URL?) -> Bool {
+    func notificationShouldDismiss(controller: BaseNotificationViewController,
+                                   callToActionURL: URL?,
+                                   shouldTrack: Bool,
```
yarneo: could we have a default value?

pchien: i think i tried to put one but since it's being used outside this file from the protocol definition, it wouldn't let me omit it so the default value wasn't useful

Mixpanel/InAppNotifications.swift
pchien: i wanted it to always send "url"

yarneo:
LGTM

Mixpanel/InAppNotifications.swift
pchien: i just tried this and it said i can't assign like that because properties is a let constant

PR #131, opened 2017-09-14
Swift 4 migration

fpg1503:
This PR is a naïve migration to Swift 4. I did the following:
- Automatic Swift 4 migration using @objc inference
- Use `SignedNumeric & Comparable` instead of `SignedNumber`
- The return of `method_getImplementation` is now `IMP` (non-Optional)
- [Force unwrap `class_getInstanceMethod(aClass, originalSelector)`](https://github.com/mixpanel/mixpanel-swift/compare/master...fpg1503:master#diff-1b0e72597ae916cee81cc2c0db9a9016L76) - I don't like this but it seems to be the best alternative -- thoughts?
- Break [these expressions](https://github.com/mixpanel/mixpanel-swift/compare/master...fpg1503:master#diff-095cae2f7c8661880e6909dab634fe3dL111) into smaller expressions that the compiler can handle (it was having trouble understanding them).
- Use [conditional cast instead of type check + force cast](https://github.com/mixpanel/mixpanel-swift/compare/master...fpg1503:master#diff-1ccd97e1344422a46dba131f9bb7f67eL244) - was there any reason for not conditionally casting before? The code seems more idiomatic now.

### DIFF
--- a/Mixpanel.xcodeproj/project.pbxproj
+++ b/Mixpanel.xcodeproj/project.pbxproj
@@ -636,7 +636,7 @@
 						CreatedOnToolsVersion = 7.3.1;
 						DevelopmentTeam = E8FVX7QLET;
 						DevelopmentTeamName = "Mixpanel, Inc.";
-						LastSwiftMigration = 0830;
+						LastSwiftMigration = 0900;
 					};
 					E12782B21D4AB4B30025FB05 = {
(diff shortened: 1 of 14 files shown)


fpg1503:
Just realized you already have a `swift4` branch, I'm closing this.

PR #135, opened 2017-09-24
Set allow-app-extension-safe-API-only

mxcl:
Otherwise generates warnings when used with Carthage.


mxcl:
ping

yarneo:
Hey @mxcl apologies for the delay. I intentionally didn't have that set to on because now the main Mixpanel package and the Mixpanel App-Extension package are part of the same target, and if we enable that flag it will possibly disallow the linking of the non-appex compatible frameworks to the main package.

mxcl:
If you set it and it doesn’t link then it means the framework is not app extension safe and your docs are wrong.

Unless I’m missing something.

yarneo:
Hi @mxcl,

If you recall correctly we had to do a workaround to get the app extension to be part of the same spec as the main framework for CocoaPods, so you could use both in your app. To do that we had to do some kind of runtime check to see if certain APIs were available and that way we knew if to invoke them (appex doesn't support some APIs). Therefore because both targets are collated if we do set that switch to ON, it could remove those APIs from the main Mixpanel framework that uses them. Whereas if we leave it as off, it doesn't hurt the Appex integration. 

If you have a smart way of turning that switch on *only* for appex usage then I'll be happy to help out and push the change.

mxcl:
>  it could remove those APIs from the main Mixpanel framework that uses them

This is not what this checkbox does. It does nothing to the resulting binary, it merely states that you support app-extensions, which you claim you do, so you should merge this PR. Thanks.

mxcl:
Imagine all the kisses and hearts spread over my posts. I forget everything sounds mean on the Internet.

You do whichever you like with this, I think this should be set based on what your project claims to support, but will handle whatever you do with it.

yarneo:
No worries at all, you contribute and help us improve our library and we are thankful for that! I just tested integrating the Mixpanel framework with allow-app-extension-safe-API-only set and it looks good. Will merge your PR, thanks!

PR #141, opened 2017-10-09
not do UI stuff on background thread

yarneo:


Mixpanel/AutomaticEvents.swift
```diff
@@ -84,7 +84,9 @@ class AutomaticEvents: NSObject, SKPaymentTransactionObserver, SKProductsRequest
                                                object: nil)
 
         SKPaymentQueue.default().add(self)
-        setupAutomaticPushTracking()
+        DispatchQueue.main.async {
```
pchien: not sure if it matters in this instance but instead of async it might be better to do something where it checks if it's currently the main thread and does sync if not. like this https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/NSThread+MPHelpers.m

yarneo: yep aware of this method, didn't see the need for the urgency here.

PR #144, opened 2017-10-10
added Core subspec for use without Decide

yarneo:


Mixpanel-swift.podspec
```diff
@@ -10,28 +10,36 @@ Pod::Spec.new do |s|
...
   s.osx.pod_target_xcconfig = {
     'OTHER_SWIFT_FLAGS' => '$(inherited) -D MAC_OS'
   }
+
+  s.subspec 'Complete' do |ss|
+    ss.ios.pod_target_xcconfig = {
+      'OTHER_SWIFT_FLAGS' => '$(inherited) -D DECIDE'
+    }
+    ss.ios.source_files = 'Mixpanel/*.swift'
+    ss.tvos.source_files = base_source_files
```
pchien: maybe leave this and osx in the default part of the spec since they're the same between the two

yarneo: I wanted to originally, but pod lib lint complained that the subspec doesn't have source_files :/

PR #153, opened 2017-10-25
added session metadata

yarneo:


Mixpanel/SessionMetadata.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import Foundation
+
+class SessionMetadata {
+    var eventsCounter: UInt64 = 0
+    var peopleCounter: UInt64 = 0
+    var sessionID: UInt64 = 0
+    var sessionStartEpoch: UInt64 = 0
+
+    func applicationDidBecomeActive() {
```
patedit: do you think we need to synchronize this? an event could be tracked before the app goes into the foreground again and counter would be 0

Mixpanel/SessionMetadata.swift
```diff
@@ -0,0 +1,43 @@
...
+        sessionStartEpoch = UInt64(Date().timeIntervalSince1970)
+    }
+
+    func toDict(isEvent: Bool = true) -> InternalProperties {
+        let dict = ["$mp_event_id": UInt64.random,
+                    "$mp_session_id": sessionID,
+                    "$mp_session_seq_id": (isEvent ? eventsCounter : peopleCounter),
+                    "$mp_session_start_sec": sessionStartEpoch]
+        isEvent ? (eventsCounter += 1) : (peopleCounter += 1)
+        print(dict)
```
patedit: remove

Mixpanel/SessionMetadata.swift
```diff
@@ -0,0 +1,43 @@
...
+        return dict
+    }
+}
+
+extension UInt64 {
+    static var random: UInt64 {
+        let hex = UUID().uuidString
```
patedit: any reason why are not md5'ng the session id with events/people counter plus an string?

Mixpanel/SessionMetadata.swift
```diff
@@ -0,0 +1,43 @@
...
+    var eventsCounter: UInt64 = 0
+    var peopleCounter: UInt64 = 0
+    var sessionID: UInt64 = 0
+    var sessionStartEpoch: UInt64 = 0
+
+    func applicationDidBecomeActive() {
+        eventsCounter = 0
+        peopleCounter = 0
+        sessionID = UInt64.random
+        sessionStartEpoch = UInt64(Date().timeIntervalSince1970)
```
patedit: i'm 99% sure this is right, but want to double check that it's in seconds and not ms

yarneo: yeah Intifying it makes it seconds as the decimal part is the milliseconds.

Mixpanel/SessionMetadata.swift
```diff
@@ -37,12 +38,8 @@ class SessionMetadata {
...
 }
 
-extension UInt64 {
-    static var random: UInt64 {
-        let hex = UUID().uuidString
-            .components(separatedBy: "-")
-            .suffix(2)
-            .joined()
-        return UInt64(hex, radix: 0x10)!
+private extension String {
```
pchien: it seems weirder to me to have this be an extension of String instead of just a function on SessionMetadata the way it is in objc. it's private anyway and if we're going to reuse it we can refactor later but for now it's just kinda confusing. feel free to disregard if you feel differently

Mixpanel/SessionMetadata.swift
```diff
@@ -11,24 +11,25 @@ import Foundation
 class SessionMetadata {
     var eventsCounter: UInt64 = 0
     var peopleCounter: UInt64 = 0
-    var sessionID: UInt64 = 0
+    var sessionID: String = ""
```
pchien: same thoughts here as objc

PR #159, opened 2017-11-24
Get rid of deprecated String.characters

gmorning:


pchien:
Thanks for this @gmorning!

PR #160, opened 2017-11-27
Call MixpanelTweaks.defaultStore.reset() in reset

jaredmixpanel:


jaredmixpanel:
Just as we did for the Obj-C lib: https://github.com/mixpanel/mixpanel-iphone/pull/653

pchien:
thanks for doing this!

PR #174, opened 2018-01-24
Update union method documentation

cherisegoode:
Update union method documentation per customer feedback.


pchien:
Thanks @cherisegoode!

PR #179, opened 2018-01-29
Added completion closures to functions that modify superProperties

lokae0:
While implementing the suggestion to restore a user's super properties described [here](https://mixpanel.com/blog/2015/09/21/community-tip-maintaining-user-identity/), we encountered a race condition where `currentSuperProperties` was still empty immediately after calling `registerSuperProperties`.

This seems to occur because `registerSuperProperties` uses the serial trackingQueue to update the `superProperties` variable, but `currentSuperProperties` must return the variable immediately.

Adding completion closures will at least allow SDK users to delay until the `superProperties` variable has been updated. Please let me know if there are any better alternative solutions. Thanks!

### DIFF
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -1102,10 +1102,13 @@ extension MixpanelInstance {
 
     /**
      Clears all currently set super properties.
+
+     - parameter completion: closure to call upon execution
      */
-    open func clearSuperProperties() {
+    open func clearSuperProperties(completion: (() -> Void)? = nil) {
         dispatchAndTrack() {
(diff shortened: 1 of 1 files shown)


davidmuzi:
Running into this same issue.  Another approach would be to make the `currentSuperProperties` enqueue synchronously.

```swift
    open func currentSuperProperties() -> [String: Any] {
		let props: [String: Any]!
		
		trackingQueue.sync {
			props = superProperties
		}
		
		return props
    }
```

This way the call sites don't need to change

lokae0:
@davidmuzi, good suggestion. I'll close this PR as it's very stale; perhaps open a new one with your change and see what the folks at Mixpanel say?

PR #188, opened 2018-03-11
Replaces fatalError with assert

kevinchau:
I had an edge case where Mixpanel was not initalized before something needed to report an event on Mixpanel - although this is my fault entirely, it would be nice if Mixpanel only crashes if a method is called on a null instance in debug.

In the event that an event is reported to mixpanel in release - use a placeholder blank instance, until a real instance with a valid token is initialized.


zihejia:
hi @kevinchau , i am sorry about the delay and thanks so much your contribution. Now we have accepted your PR.

PR #190, opened 2018-03-22
Add iOS IFA to default event+people properties

ryanseams:
Adds the iOS IFA to default event+people properties, which establishes consistency with the Objective-C library. This code is modeled off of the code in `MixpanelInstance.swift` for setting IFA as Distinct ID.


zihejia:
Thank you for providing the PR. Really sorry about the fact we no longer providing support for IFA.

Issue #213, opened 2018-06-25
Flushing event queue not sending multiple batches


erenliving:
It appears the check for remaining events in the shadowQueue copy is preventing multiple batches of events to be flushed when the event queue has more than APIConstants.batchSize events in it, instead it is removing all events from the queue after the first batch is sent successfully. This causes the remaining queued events to be lost.

zihejia:
Thanks @erenliving for raising this issue and providing a fix for this! We will release a new version.

zihejia:
v2.4.4 has been released and contains the fix, once again thanks @erenliving !

PR #214, opened 2018-06-25
Flush all queued events in batches (#1)

erenliving:
The check for removing events from the queue after a successful flushRequest is removing all events from the shadowQueue. When queue.count is greater than APIConstants.batchSize, it will cause the first batch of events up to the batchSize to be sent and the remaining events in the queue to be dropped.


erenliving:
Fixes issue #213

PR #219, opened 2018-07-15
Store mixpanel files inside "Application Support/Mixpanel" on iOS

phoney:
Apple best practice for iOS  is that support files are stored in subfolders inside the "Application Support" folder in the app sandbox. The code in this pull request will run on iOS only.

New installations will create the "Application Support/Mixpanel" folder and store all files there. Existing installations will create the "Application Support/ Mixpanel" folder and move existing "mixpanel-" files there.


phoney:
@zihejia Can you review this PR and merge?

zihejia:
we will refactor the offline storage, thanks for providing the PR.

PR #244, opened 2018-09-26
fix notification tracking crash related to @noescape

Macmee:
the `typealias` in the swizzling block was missing `@escaping`


zihejia:
Hi @Macmee , thanks for the PR.

squarelab-jwlee:
@Macmee Great work!! I think this is the reason of my crash on Xcode10. But on line 289, it defines type alias for another `MyCFunction` and it uses closure `() -> Void` just like this code. I think my crash occurs there. I think it needs `@escaping` attribute, too to fix my crash. May I pull-request for that?

zihejia:
hi @squarelab-jwlee , you are more than welcome to provide a pull-request. Thanks so much!

Issue #245, opened 2018-09-26
"trackPushNotification is inaccessible due to 'internal' protection level"


mansimransingh:
`trackPushNotification` needs to be marked as `open` for it to be accessible.

https://github.com/mixpanel/mixpanel-swift/blob/ca7f490e1479f5b51411a35b4ca3f10b8d55da5f/Mixpanel/MixpanelInstance.swift#L1095

zihejia:
Hi @mansimransingh ,  sorry for the late response, `trackPushNotification` is designed to be called automatically internally, so there is no need to called explicitly for users. Is there any case you need to call directly?

mansimransingh:
Hey @zihejia, if I understand correctly, the automated tracking only works when the app is launched from a push notification from a cold start, and relies on `UIApplicationLaunchOptionsKey.remoteNotification`. 

I'd also like to track push notifications received while app is running in foreground/background. To achieve that we manually call `trackPushNotification` in two other places.

Our `UIApplicationDelegate`

```
func application(_ application: UIApplication, 
didReceiveRemoteNotification userInfo: [AnyHashable : Any], 
   fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void)
```

and in our `UNUserNotificationCenterDelegate`.

zihejia:
hi @mansimransingh , it should work for all the cases you mentioned automatically unless you need to use a custom event name. In `AutomaticEvents.swift`, we swizzle those delegate methods and embed `trackPushNotification`

fphilipe:
> `trackPushNotification` is designed to be called automatically internally, so there is no need to called explicitly for users. Is there any case you need to call directly?

Yes. We have set `automaticPushTracking` to `false` because we don't want any SDKs performing method swizzling in general.

That doesn't mean that we don't want to track push notifications. We've had to copy the implementation of [`trackPushNotification`](https://github.com/mixpanel/mixpanel-swift/blob/d8b079a5cb6c899c56b36a9b2b9756f9f3cf90dc/Mixpanel/MixpanelInstance.swift#L1276-L1301) into our app due to it being inaccessible.

Would it be possible to make that method public?

PR #246, opened 2018-09-26
Mark trackPushNotification as open so that it is accessible outside t…

mansimransingh:
…he framework


mansimransingh:
Fixes [`245`](https://github.com/mixpanel/mixpanel-swift/issues/245)

PR #256, opened 2018-10-08
Fix crash on tapping notification when built by Xcode 10

squarelab-jwlee:
Add @escaping attribute on closure type argument of typealiased MyCFunction


zihejia:
Thanks @squarelab-jwlee !

PR #258, opened 2018-10-11
Unset VALID_ARCH setting

krbarnes:
Building apps for simulator results in this error right now. Removing the VALID_ARCH build setting allows the builds to succeed again.

```
Mapping architecture arm64 to x86_64. Ensure that this target's Architectures and Valid Architectures build settings are configured correctly for the iOS Simulator platform. (in target 'Mixpanel')
```


zihejia:
Hi @krbarnes , thanks for providing the PR!

PR #273, opened 2018-12-14
Issue 272: Handle archiveToFile exception and log the issue

drfindley:
This uses the Obj-C ExceptionWrapper to catch the exception when trying to use NSKeyedArchiver. This also logs what happened so we can see what the issue is.


zihejia:
Thanks a lot @drfindley , sorry for the delay.

PR #281, opened 2019-01-22
Optimize `.reduce` usage on Array

asiliuk:
✋ Just a small fix that can improve `.reduce` performance
In previous implementation it would copy array on every iteration
In new one only one instance is created and `.append` is just linear operation
So it should go from `O(n^2)` -> `O(n)`

### DIFF
--- a/Mixpanel/TweakCollection.swift
+++ b/Mixpanel/TweakCollection.swift
@@ -34,8 +34,8 @@ extension TweakCollection {
 	/// A flat list of all the tweaks in a TweakCollection.
 	/// Used for sharing out the contents of a TweakStore.
 	internal var allTweaks: [AnyTweak] {
-		return sortedTweakGroups.reduce([]) {
-			$0 + $1.sortedTweaks
+		return sortedTweakGroups.reduce(into: []) {
+			$0.append(contentsOf: $1.sortedTweaks)
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


Mixpanel/TweakCollection.swift
```diff
@@ -34,8 +34,8 @@ extension TweakCollection {
 	/// A flat list of all the tweaks in a TweakCollection.
 	/// Used for sharing out the contents of a TweakStore.
 	internal var allTweaks: [AnyTweak] {
-		return sortedTweakGroups.reduce([]) {
-			$0 + $1.sortedTweaks
+        return sortedTweakGroups.reduce(into: []) {
```
zihejia: please add the correct indentation

Mixpanel/TweakStore.swift
```diff
@@ -119,10 +119,11 @@ public final class TweakStore {
 		persistence.clearAllData()
 
 		// Go through all tweaks in our library, and call any bindings they're attached to.
-		tweakCollections.values.reduce([]) { $0 + $1.sortedTweakGroups.reduce([]) { $0 + $1.sortedTweaks } }
-			.forEach { updateBindingsForTweak($0)
-		}
-
+        tweakCollections.values
```
zihejia: please add the correct indentation

zihejia:
thanks @asiliuk ,  looks good except the code formatting.

zihejia:
thanks so much for your contribution @asiliuk !

PR #287, opened 2019-01-31
Add support for display triggers in inapp notifications

madster26:
**Event Triggered InApps**

Currently users have no control over when an InApp notification shows up. With this release users can now control when an InApp gets displayed based on an event that they track within their app. This "trigger" event is defined during message creation. You can additionally filter the event based on properties that are tracked along with the event for even finer controls.

### DIFF
--- a/Mixpanel/DisplayTrigger.swift
+++ b/Mixpanel/DisplayTrigger.swift
@@ -1,0 +1,59 @@
+//
+//  DisplayTrigger.swift
+//  Mixpanel
+//
+//  Created by Madhu Palani on 1/30/19.
+//  Copyright © 2019 Mixpanel. All rights reserved.
+//
+
+import Foundation
+
+class DisplayTrigger {
+    enum PayloadKey {
+        static let event = "event"
+        static let selector = "selector"
+    }
+    
+    static let ANY_EVENT = "$any_event"
+    
+    let event: String?
+    let selector: [String: Any]?
+    
+    init?(JSONObject: [String: Any]?) {
+        guard let object = JSONObject else {
+            Logger.error(message: "display trigger json object should not be nil")
+            return nil
+        }
+        
+        guard let event = object[PayloadKey.event] as? String else {
+            Logger.error(message: "invalid event name type")
+            return nil
+        }
+        
+        guard let selector = object[PayloadKey.selector] as? [String: Any]? else {
+            Logger.error(message: "invalid selector section")
+            return nil
+        }
+        
+        self.event = event
+        self.selector = selector
+    }
+    
+    func matchesEvent(eventName: String?, properties: Properties? = nil) -> Bool {
+        if let event = self.event {
+            if (eventName == DisplayTrigger.ANY_EVENT || eventName == "" || eventName?.caseInsensitiveCompare(event) == ComparisonResult.orderedSame) {
+                return true
+            }
+        }
+        
+        return false
+    }
+    
+    func payload() -> [String: AnyObject] {
+        var payload = [String: AnyObject]()
+        payload[PayloadKey.event] = event as AnyObject
+        payload[PayloadKey.selector] = selector as AnyObject
+        
+        return payload
+    }
+}
--- a/Mixpanel/DisplayTrigger.swift
+++ b/Mixpanel/DisplayTrigger.swift
@@ -17,10 +17,10 @@ class DisplayTrigger {
     static let ANY_EVENT = "$any_event"
     
     let event: String?
-    let selector: [String: Any]?
+    let selectorOpt: [String: Any]?
     
-    init?(JSONObject: [String: Any]?) {
-        guard let object = JSONObject else {
+    init?(jsonObject: [String: Any]?) {
+        guard let object = jsonObject else {
             Logger.error(message: "display trigger json object should not be nil")
             return nil
         }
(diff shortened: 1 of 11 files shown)


Mixpanel/InAppNotification.swift
```diff
@@ -55,7 +57,23 @@ class InAppNotification {
...
-            Logger.error(message: "invalid notification extra section")
+            Logger.error(message: "invalid notification extra section \(object)")
+            return nil
+        }
+        
+        if let rawDisplayTriggers = object[PayloadKey.displayTriggers] as? [[String: Any]]? {
+            if let rawDisplayTriggers = rawDisplayTriggers {
+                for rawDisplayTrigger in rawDisplayTriggers {
+                    if let displayTrigger = DisplayTrigger(JSONObject: rawDisplayTrigger) {
+                        self.displayTriggers.append(displayTrigger)
```
zihejia:
```suggestion
                        displayTriggers.append(displayTrigger)
```

zihejia: Code style: in swift, preferably not adding self. unless necessary

Mixpanel/InAppNotification.swift
```diff
@@ -110,6 +145,11 @@ class InAppNotification {
         payload[PayloadKey.bodyColor] = bodyColor as AnyObject
         payload[PayloadKey.type] = type as AnyObject
         payload[PayloadKey.body] = body as AnyObject
+        var pldDisplayTriggers = [[String: AnyObject]]()
```
zihejia: what is `pld`, any better name to be used?

Mixpanel/DisplayTrigger.swift
```diff
@@ -0,0 +1,69 @@
...
+        static let selector = "selector"
+    }
+    
+    static let ANY_EVENT = "$any_event"
+    
+    let event: String?
+    let selector: [String: Any]?
+    let selectorEvaluator: SelectorEvaluator?
+    
+    init?(JSONObject: [String: Any]?) {
```
zihejia: nit: jsonObject,

Mixpanel/InAppNotification.swift
```diff
@@ -55,7 +57,23 @@ class InAppNotification {
         }
 
         guard let extras = object[PayloadKey.extras] as? [String: Any] else {
-            Logger.error(message: "invalid notification extra section")
+            Logger.error(message: "invalid notification extra section \(object)")
+            return nil
+        }
+        
+        if let rawDisplayTriggers = object[PayloadKey.displayTriggers] as? [[String: Any]]? {
+            if let rawDisplayTriggers = rawDisplayTriggers {
```
zihejia: is it redundant here for assigning `rawDisplayTriggers` again?

Mixpanel/SelectorEvaluator.swift
```diff
@@ -0,0 +1,604 @@
...
+    static let LESS_THAN_OPERATOR = "<"
+    static let LESS_THAN_EQUAL_OPERATOR = "<="
+    // Unary operators
+    static let NOT_OPERATOR = "not"
+    static let DEFINED_OPERATOR = "defined"
+    static let NOT_DEFINED_OPERATOR = "not defined"
+    // Special words
+    static let NOW_LITERAL = "now"
+    
+    let selector: [String:Any]?
```
zihejia:
```suggestion
    let selector: [String: Any]?
```

Mixpanel/SelectorEvaluator.swift
```diff
@@ -0,0 +1,604 @@
...
+    
+    init?(JSONObject: [String: Any]?) {
```
zihejia: nit: 
```suggestion
    init?(jsonObject: [String: Any]?) {
```
The same as above

Mixpanel/SelectorEvaluator.swift
```diff
@@ -0,0 +1,604 @@
...
+        guard let op = node[OPERATOR_KEY] as? String, op == NUMBER_OPERATOR else {
+            Logger.error(message: "invalid operator: number")
+            return nil
+        }
+        guard let children = node[CHILDREN_KEY] as? [[String: Any]], children.count == 1 else {
+            Logger.error(message: "invalid operator: number")
+            return nil
+        }
+        
+        return toNumber(value: evaluateNode(node: children[0], properties: properties))
```
zihejia: would be better to define const for `0` and `1` and avoid using magic number in this case

Mixpanel/SelectorEvaluator.swift
```diff
@@ -0,0 +1,604 @@
...
+        }
+        
+        guard let unit = window[UNIT_KEY] as? String else {
+            Logger.error(message: "missing unit")
+            return nil
+        }
+        let date = getCurrentDate()
+        switch unit {
+        case HOUR_KEY:
+            return date.addingTimeInterval(value.doubleValue*60*60)
```
zihejia: nit: value.doubleValue * 60 * 60

Mixpanel/SelectorEvaluator.swift
zihejia: can SelectorEvaluator even be a stateless class so you don't even need `init`?

Mixpanel/InAppNotification.swift
madster26: rawDisplayTriggers is optional let me rename it to make it obvious

PR #288, opened 2019-02-26
Allow to remove All device push tokens

asiliuk:
Hi!

I just want to bring missing function from `mixpanel-iphone` to swift version


zihejia:
Thank you @asiliuk for providing this.

PR #289, opened 2019-03-20
Prefix swizzled method name

ChrisGriffiths:
When swizzling ensure the category method are prefix with your framework since it has global scope

We are experiencing an issue where 2 SDK have exactly the same method signature causing a crash. Prefixing the swizzled methods with `mp_` resolves the issue something that is adviced to do when swizzling https://nshipster.com/method-swizzling/


zihejia:
Thanks @ChrisGriffiths for providing this

Issue #97, opened 2017-06-25
All events are not sending


wenxin3262:
Enabled debug logging and see this all over the place:
`[Mixpanel - JSONHandler.swift - func serializeJSONObject] (warning) - object isn't valid and can't be serialzed to JSON`

Mixpanel-swift version: 2.1.6 - 2.1.8
XCode Version: 8.3.3

problem: If there is one event with incorrect props all queued events will not be sent. The expected behavior is ignoring that bad event and send all others.

yarneo:
Hey @wenxin3262 , to help me diagnose the issue, could you please show me the object it is trying to serialize so I can understand why it isn't able to serialize it into a JSON Object?

One way you could do this is issue a `print` or put a debug point where `guard JSONSerialization.isValidJSONObject(serializableJSONObject) else {` is called. This is in the file JSONHandler.swift line 37. I would try to see what the value of `serializableJSONObject` is and what the value of `obj` is. This would be very helpful, thanks!

yarneo:
hey @wenxin3262 I've tried to see how it is possible to make the object not serializable and I'm having a hard time seeing it. The reason being is that prior to calling `JSONSerialization.isValidJSONObject` we call `makeObjectSerializable` on the object, which basically enforces all "unserializable" objects to a String, and let's serializable objects pass through. I am guessing the only way it could not be a valid JSON object if somehow the top level of `obj` is not an Array or Dictionary, which would be impossible as a batch of events is always sent as an Array. Therefore I would love to get an idea of what object you are sending that is causing this so we could remedy this for future iterations. Feel free to open this ticket once you have more data to share. Thanks!

wenxin3262:
@yarneo My point is that once JSONSerialization failed and all other queued events will get stuck, the client will never send another event. To make it fail just do something stupid like divide by zero.

yarneo:
Hey @wenxin3262 totally understand your point. The reason I tried to get more information from you on how you got to that is because we call `makeObjectSerializable` before that converts all the "bad unserializable" objects to serializable ones. If somehow one of the objects passed through that and still you got `object isn't valid and can't be serialized to JSON` then it would be great to know what it is. In theory, this warning should never happen.

PR #303, opened 2019-05-28
Missing case in makeObjectSerializable() : NaN or infinity cannot be serialized as JSON.

tonysung:
#97 is a legitimate issue. This is a proper fix.

### DIFF
--- a/Mixpanel/JSONHandler.swift
+++ b/Mixpanel/JSONHandler.swift
@@ -50,6 +50,10 @@ class JSONHandler {
 
     private class func makeObjectSerializable(_ obj: MPObjectToParse) -> MPObjectToParse {
         switch obj {
+        case let obj as Double where !obj.isFinite:
+            Logger.info(message: "enforcing string on non-finite numbers")
+            return String(describing: obj)
+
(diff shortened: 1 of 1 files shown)


tonysung:
@zihejia pls review and merge.

Mixpanel/JSONHandler.swift
```diff
@@ -50,6 +50,10 @@ class JSONHandler {
 
     private class func makeObjectSerializable(_ obj: MPObjectToParse) -> MPObjectToParse {
         switch obj {
+        case let obj as Double where !obj.isFinite:
+            Logger.info(message: "enforcing string on non-finite numbers")
+            return String(describing: obj)
+
```
zihejia: Thanks @tonysung ! One small nit suggestion would be as the following, let the exception handled by the default.

           case let obj as Double where obj.isFinite:
            return obj
            
           case is String, is Int, is UInt, is UInt64, is Float, is Bool:
            return obj

zihejia:
Thanks @tonysung，I am merging it now, will tweak it a tiny bit later.

PR #328, opened 2019-09-04
Move from ObjC locks to async dispatch on serial queue

RamblinWreck77:
Notes: 
-Serial queues have a guarantee from the OS to be run sequentially, so we can safety async dispatch from any thread to archiveQueue and our operations will run with a safe barrier between them in the order they come in. This should achieve the same behavior as before in a more efficient (and hopefully safe) manner.

-Renamed archiveToFile > _archiveToFile, and added a comment indicating that said function should only be called inside a dispatch to archiveQueue. archiveQueue's QoS was also set to "utility", as 3rd party libraries like mixpanel should be doing as little work on the main thread as possible. The .utility QoS will ensure optimal thread assignment.

-archiveQueue dispatch was set as .async, as .sync is always risky and I didn't see any specific reason to force sync. Will test this w/our QA and see if anything comes up.

### DIFF
--- a/Mixpanel/Persistence.swift
+++ b/Mixpanel/Persistence.swift
@@ -26,6 +26,9 @@ struct ArchivedProperties {
 
 class Persistence {
 
+    private static let archiveQueue: DispatchQueue = DispatchQueue(label: "com.mixpanel.archiveQueue",
+                                                                   qos: .utility)
+
     enum ArchiveType: String {
         case events
         case people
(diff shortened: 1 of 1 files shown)


RamblinWreck77:
Hmmm... test failures. That's odd. I'll figure this out real quick.

RamblinWreck77:
Ok, I got all tests to pass locally. Turns out it has to use .sync... disappointing from a performance perspective but it should be safe since this queue is private to the Persistence class. Updating now.

RamblinWreck77:
Hmmmm... That's odd. Passed all tests locally.

RamblinWreck77:
@zihejia Can you check this out? It passes all of the tests locally in xcode, I suspect there is a config issue on the CI.

This one is particularly weird:

```
testArchive, XCTAssertEqual failed: ("d1") is not equal to ("11C131D2-EB7B-4E50-B241-033228D39FF2") - default distinct id archive failed
```

zihejia:
Hi @RamblinWreck77 , i have opened a PR to address this issue https://github.com/mixpanel/mixpanel-swift/pull/330, it's very close to your one with just a few tweaks. As always, thanks so much for your contribution. I will close this one now.

Issue #327, opened 2019-09-04
[2.6.5] Crash in archiveToFile's NSKeyedArchiver


RamblinWreck77:
Hello! We're getting a weird production crash inside Persistence.swift's archiveToFile function. 

It seems like something is blowing up inside NSKeyedArchiver.archiveRootObject's, and I can't reproduce it but we're seeing ~100 crashes a day from it. 

```
Sampled Backtrace (0)#0	(null) in bufferWrite ()
#1	(null) in __CFBinaryPlistWriteOrPresize ()
#2	(null) in -[NSKeyedArchiver finishEncoding] ()
#3	(null) in +[NSKeyedArchiver archiveRootObject:toFile:] ()
#4	0x3409fa8 in closure #1 in static Persistence.archiveToFile(_:object:token:) 
#5	(null) in thunk for @escaping @callee_guaranteed () -> () ()
#6	0x3385e58 in +[ExceptionWrapper try:catch:finally:] 
#7	0x340d8f0 in specialized static Persistence.archiveToFile(_:object:token:)
#8	(null) in closure #2 in closure #1 in MixpanelInstance.track(event:properties:) ()
#9	0x340e7e0 in closure #1 in ReadWriteLock.read(closure:)
#10	(null) in partial apply for thunk for @callee_guaranteed () -> () ()
#11	(null) in thunk for @escaping @callee_guaranteed () -> () ()
#12	(null) in _dispatch_client_callout ()
#13	(null) in _dispatch_sync_invoke_and_complete ()
#14	0x33da058 in closure #1 in MixpanelInstance.track(event:properties:)
#15	(null) in thunk for @escaping @callee_guaranteed () -> () ()
#16	(null) in _dispatch_call_block_and_release ()
#17	(null) in _dispatch_client_callout ()
#18	(null) in _dispatch_lane_serial_drain ()
#19	(null) in _dispatch_lane_invoke ()
#20	(null) in _dispatch_workloop_worker_thread ()
#21	(null) in _pthread_wqthread ()

```

Taking a peek at the source code it's not obvious to me what the problem is, but I'm suspicious of all the objc_sync_enter/exit mixed with ExceptionWrapper.try and archiveRootObject. 

I think I see what you're trying to do here with the sync locks, would you consider switching to a more modern serial queue + async dispatch model? That would get rid of the need for locks and *might* fix the issue. I'll get a PR up and reference here and see if you like it.

RamblinWreck77:
#328 up

[...]

PR #330, opened 2019-10-03
Xcode 11 support and race condition fixes

zihejia:
Fix issues: 
1. Fix a rare case where a race case could happen for Logger
2. Fix the race condition/crash for `archive`

Related issue:
https://github.com/mixpanel/mixpanel-swift/issues/327

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
   - gem install fourflusher
   - gem uninstall cocoapods --force -a -q -I -x
   - gem install cocoapods -v 1.6.1
-osx_image: xcode10.2
+osx_image: xcode11
 before_script: cd MixpanelDemo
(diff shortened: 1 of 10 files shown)


jserrafindster:
this seems to keep happening to our users with iOS 10 iphone 5

zihejia:
hi @jserrafindster, thanks for bring it up. Is it only happening to iphone 5c users? Do you mind open an issue with more details?

jserrafindster:
> hi @jserrafindster, thanks for bring it up. Is it only happening to iphone 5c users? Do you mind open an issue with more details?

Seems to happen on iphone5 variations and ipad 4 i created an issue here
[https://github.com/mixpanel/mixpanel-swift/issues/358](https://github.com/mixpanel/mixpanel-swift/issues/358)

PR #338, opened 2020-01-09
fix in app notification not showing or showing correctly in iOS 13

zihejia:
- Fix mini notification not showing properly in some devices.
- Fix In-apps not showing up for iOS 13 with Scenes.

### DIFF
--- a/Mixpanel/TakeoverNotificationViewController.swift
+++ b/Mixpanel/TakeoverNotificationViewController.swift
@@ -140,10 +140,22 @@ class TakeoverNotificationViewController: BaseNotificationViewController {
     }
 
     override func show(animated: Bool) {
-        window = UIWindow(frame: CGRect(x: 0,
-                                        y: 0,
-                                        width: UIScreen.main.bounds.size.width,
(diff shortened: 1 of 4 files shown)


Mixpanel/TakeoverNotificationViewController.swift
```diff
@@ -140,10 +140,22 @@ class TakeoverNotificationViewController: BaseNotificationViewController {
     }
 
     override func show(animated: Bool) {
-        window = UIWindow(frame: CGRect(x: 0,
-                                        y: 0,
-                                        width: UIScreen.main.bounds.size.width,
-                                        height: UIScreen.main.bounds.size.height))
+#if !BUILDING_FOR_APP_EXTENSION
+        if #available(iOS 13.0, *) {
```
jbwyme: will this still work in iOS13 if they _aren't_ using scenes?

zihejia: yes, it works. iOS 13 replace the top level UIWindow to UIWindowScene regardless whether you use SceneDelegate or not.

PR #340, opened 2020-01-14
Add support for rich push notifications

jbwyme:


Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,125 @@
...
+        self.maybeAttachButtons(bestAttemptContent: bestAttemptContent) {
+            self.maybeAttachMedia(bestAttemptContent: bestAttemptContent) {
+                contentHandler(bestAttemptContent)
+            }
+        }
+    }
+    
+    func maybeAttachButtons(bestAttemptContent: UNMutableNotificationContent, completionHandler: @escaping () -> Void) {
+        guard let buttons = bestAttemptContent.userInfo["mp_buttons"] as? [[AnyHashable: Any]] else {
+            NSLog("maybeAttachButtons: No action buttons found in the push notification payload.")
```
jbwyme: I realize that `NSLog` might not be the correct way to log this debugging info but trying to use `print` or the `MPLog` stuff resulted in nothing showing up. `NSLog` was the only way I was able to see output in the console.

Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,168 @@
...
+
+    var contentHandler: ((UNNotificationContent) -> Void)?
+    var bestAttemptContent: UNMutableNotificationContent?
+
+    open override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+
+        NSLog("%@ MPNotificationServiceExtension didReceiveNotificationRequest", self);
+
+        guard MixpanelPushNotifications.isMixpanelPushNotification(request.content) else {
+            NSLog("%@ Not a Mixpanel push notification, returning original content", self);
```
zihejia: Can we use Logger.info in here?

jbwyme: No, my understanding is that this is a completely new environment which means it's a different Mixpanel instance. Thus, we'd have to require customers to explicitly create a Mixpanel instance here and enable logging to get these. My thoughts were that this is a pretty rare thing so always `NSLog`ing won't pollute the logs much.

Mixpanel/PushNotifications.swift
```diff
@@ -0,0 +1,149 @@
...
+                guard let buttonOnTap = button["ontap"] as? [AnyHashable: Any] else {
+                    Logger.debug(message: "Expected 'ontap' property in button dict")
+                    completionHandler();
+                    return
+                }
+
+                ontap = buttonOnTap
+
+                trackingProps["tap_target"] = "button"
+                trackingProps["button_id"] = button["id"] as! String
```
zihejia: prefer safely unwrap the optional in here

Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,168 @@
...
+        let buttons = content.userInfo["mp_buttons"] as? [[AnyHashable: Any]] ?? []
+        if (buttons.count == 0) {
+            NSLog("getCategoryIdentifier: No action buttons found in the push notification payload.")
+        }
+        
+        // Build a list of actions from the buttons data
+        var actions: [UNNotificationAction] = []
+        for (idx, button) in buttons.enumerated() {
+            let identifier = String(format: "MP_ACTION_%lu", idx)
+            let title = button["lbl"] as! String
```
zihejia: prefer safely unwrap the optional in here

Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,168 @@
...
+        self.getCategoryIdentifier(content: request.content) { categoryIdentifier in
+
+            if let categoryIdentifier = categoryIdentifier {
+                NSLog("Using categoryIdentifer: \(categoryIdentifier)")
+                bestAttemptContent.categoryIdentifier = categoryIdentifier
+            }
+
+            // Download rich media and create an attachment
+            self.buildAttachments(content: request.content) { attachments in
+                if attachments != nil {
```
zihejia:
```                if let attachments = attachments {
                    NSLog("Adding \(attachments.count) attachment(s)")
                    bestAttemptContent.attachments = attachments
                }
```

Mixpanel/PushNotifications.swift
```diff
@@ -0,0 +1,149 @@
...
+            return;
+        }
+
+        var ontap: [AnyHashable: Any]? = nil
+
+        if response.actionIdentifier == UNNotificationDefaultActionIdentifier {
+            // The action that indicates the user opened the app from the notification interface.
+            trackingProps["tap_target"] = "notification";
+
+            if (userInfo["mp_ontap"] != nil) {
```
zihejia: nit: if let ..., 
so you don't have to force wrap.

MixpanelDemo/NotificationServiceExtension/NotificationService.swift
```diff
@@ -0,0 +1,12 @@
...
+//  NotificationService.swift
+//  NotificationServiceExtension
+//
+//  Created by Josh Wymer on 1/18/20.
+//  Copyright © 2020 Mixpanel. All rights reserved.
+//
+
+import Mixpanel
+
+class NotificationService: MixpanelNotificationServiceExtension  {}
```
zihejia: is this class being used?

Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,168 @@
+import UserNotifications
+
+@available(iOS 11.0, *)
+open class MixpanelNotificationServiceExtension: UNNotificationServiceExtension {
+
+    var contentHandler: ((UNNotificationContent) -> Void)?
+    var bestAttemptContent: UNMutableNotificationContent?
+
+    open override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+
```
zihejia: nit, remove empty line

Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        // Store a reference to the mutable content and the contentHandler on the class so we
+        // can use them in serviceExtensionTimeWillExpire if needed
+        self.contentHandler = contentHandler
+        self.bestAttemptContent = bestAttemptContent
+
+        // Setup the category first since it's faster and less likely to cause time to expire
+        self.getCategoryIdentifier(content: request.content) { categoryIdentifier in
+
```
zihejia: nit, remove empty line

Mixpanel/MixpanelNotificationServiceExtension.swift
```diff
@@ -0,0 +1,168 @@
...
+        guard let bestAttemptContent = self.bestAttemptContent else {
+            return;
+        }
+
+        contentHandler(bestAttemptContent);
+    }
+    
+    func getCategoryIdentifier(content: UNNotificationContent, completionHandler: @escaping (String?) -> Void) {
+
```
zihejia: nit, remove empty line

Mixpanel/PushNotifications.swift
jbwyme: we don't actually force unwrap in this instance

PR #378, opened 2020-06-15
Additional safety when reading device radio props

RamblinWreck77:
1) Use for-where pattern instead of for -> nested if statement to be more swifty

2) Remove need to cast to NSString

3) Breakdown value strings by components, vs unsafe .substring

4) Verify component count, else break this iteration and continue the loop


Mixpanel/MixpanelInstance.swift
```diff
@@ -674,12 +674,22 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
...
+                    // Something went wrong and we have more than prefix:target
+                    guard components.count == 2 else {
+                        continue
                     }
+
+                    // Safe to directly access by index since we confirmed count == 2 above
+                    let radioValue = components[1]
+                    
+                    // Send to parent
+                    radio = radio.count > 0 ? ", \(radioValue)" : radioValue
```
zihejia, line 690: sorry my bad from the originally code. it should be 
```suggestion
                    radio += radio.count > 0 ? ", \(radioValue)" : radioValue
```

PR #389, opened 2020-07-08
Remove using IFA as distinct id

zihejia:
- After iOS 14, accessing IFA will require an app developer to obtain an end-user's consent via an opt-in. An app will need to obtain permission to access this unique identifier, it imposes some challenges for us like caching, permissions and usage of new frameworks and doesn't add a lot of value.  We will default to random UUID instead. If you want to use a unique persistent UUID(IFV), you can define the MIXPANEL_UNIQUE_DISTINCT_ID in your  <code>Active Compilation Conditions</code> build settings. 
- Customers can still set/identify the distinct_id as IFA or `$ios_ifa property` by themselves if they have access to the advertising identifier.

### DIFF
--- a/Mixpanel/MixpanelInstance.swift
+++ b/Mixpanel/MixpanelInstance.swift
@@ -614,7 +614,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
         #elseif os(OSX)
         let distinctId = MixpanelInstance.macOSIdentifier()
         #endif
-        #else // use a random UUID by default 
+        #else // use a random UUID by default
         let distinctId: String? = UUID().uuidString
         #endif
         guard let distId = distinctId else {
@@ -725,7 +725,7 @@ extension MixpanelInstance {
     /**
      Sets the distinct ID of the current user.
 
-     Mixpanel use a random persistent UUID as the default local distinct ID.
+     Mixpanel uses a randomly generated persistent UUID  as the default local distinct ID.
 
      If you want to  use a unique persistent UUID, you can define the
      <code>MIXPANEL_UNIQUE_DISTINCT_ID</code> flag in your <code>Active Compilation Conditions</code>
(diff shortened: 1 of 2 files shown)


Mixpanel/MixpanelInstance.swift
```diff
@@ -755,26 +725,14 @@ extension MixpanelInstance {
...
-     build settings, and Mixpanel will use the IFV as the default local distinct ID.
-
-     If we are unable to get an IFA or IFV, we will fall back to generating a
-     random persistent UUID. If you want to always use a random persistent UUID
-     you can define the <code>MIXPANEL_RANDOM_DISTINCT_ID</code> preprocessor flag
-     in your build settings.
+     Mixpanel use a random persistent UUID as the default local distinct ID.
+
+     If you want to  use a unique persistent UUID, you can define the
+     <code>MIXPANEL_UNIQUE_DISTINCT_ID</code> flag in your  <code>Active Compilation Conditions</code>
```
patedit, line 731:
```suggestion
     <code>MIXPANEL_UNIQUE_DISTINCT_ID</code> flag in your <code>Active Compilation Conditions</code>
```

patedit: Plus same comment from objective-c

Mixpanel/MixpanelInstance.swift
```diff
@@ -605,55 +605,25 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
...
+        #if !os(OSX) && !WATCH_OS
+        var distinctId: String? = nil
+        if NSClassFromString("UIDevice") != nil {
             distinctId = UIDevice.current.identifierForVendor?.uuidString
         }
         #elseif os(OSX)
         let distinctId = MixpanelInstance.macOSIdentifier()
-        #else
-        let distinctId: String? = nil
         #endif // os(OSX)
```
patedit, line 616: update comment?

Issue #320, opened 2019-08-04
SwiftPM support


michael-mckenna:
Hi,
  [SwiftPM](https://goo.gl/E01ufX) is Swift's official package manager.
  
  Here some articles about SwiftPM:
  
  * [An Introduction to the Swift Package Manager](https://goo.gl/byiOc6)
  * [Swift Dependency Management for iOS](https://goo.gl/7eaz1A)
  * [Getting Started with Swift Package Manager (SwiftPM)](https://goo.gl/rCHeQU)
  * [Swift.org: Package Manager](https://goo.gl/E01ufX)
  
  When your component is compatible with SwiftPM add on your `README.md` [![SwiftPM Compatible](https://camo.githubusercontent.com/52e3db230991dda10458295d71c859c34be466b6/68747470733a2f2f696d672e736869656c64732e696f2f62616467652f5377696674504d2d436f6d70617469626c652d627269676874677265656e2e737667)](https://goo.gl/E01ufX) badge on top and add a section for the installation with SwiftPM.

gabrielPeart:
Super important to add support for Swift Package Manager

zihejia:
Thanks for bringing it up. We will add the SwiftPM support very soon.

JonathanDowning:
Hey @zihejia! 
I was just wondering how SPM support is progressing, have you run into any problems with resources such as storyboard/xib files?

RamblinWreck77:
@zihejia Just checking in, we would love this as well. We're trying to migrate away from Cocoapods to Xcode 11's SPM and mixpanel is the last 3rd party library without support. It appears #325 solves the issue.

emisvx:
Any updates on it? We are also trying to migrate from Carthage to SPM, we would really appreciate if Mixpanel added support for it.

Thanks!

RamblinWreck77:
Any update? This seems like a very minor change (adding 1 swift package file).

rodrigoelp:
I am going to join the petition to have SPM support soon-ish.

geertbeskers:
Is there any update? I'm also waiting for SPM support! Its just a minor implementation.

mdoghri:
+1

gabrielPeart:
+1

gsabran:
+1! @zihejia do you have a sense of when we can expect an update on this?

zihejia:
hi @gsabran We will release it before the Christmas break.

rodrigoelp:
[...]

PR #392, opened 2020-07-09
Add support for SPM

gsabran:
See https://github.com/mixpanel/mixpanel-swift/issues/320

I only tested for iOS. I also removed some ObjC files as SPM doesn't support this.


gsabran:
So great, thanks for the hardwork!

PR #411, opened 2020-10-15
Add Swift Package Manager support

zihejia:
This is to add Swift Package Manager support.   [SwiftPM](https://goo.gl/E01ufX) is Swift's official package manager.

How to use Swift Package Manager to integrate Mixpanel(available Xcode 12 and forward):
In Xcode, select File > Swift Packages > Add Package Dependency.
Follow the prompts using the URL for this repository and a minimum semantic version of v2.8.0

Important: Currently only works in iOS. We will soon add support for all supported platforms.


Sources/AutomaticProperties.swift
```diff
@@ -117,8 +117,8 @@ class AutomaticProperties {
     }
     #endif
 
-    class func libVersion() -> String? {
-        return Bundle(for: self).infoDictionary?["CFBundleShortVersionString"] as? String
+    class func libVersion() -> String {
+        return "2.7.7"
```
jaredmixpanel, line 121: So this is going to be a hardcoded string here now?

jaredmixpanel (approved):
lgtm, only one small question just to confirm

Sources/AutomaticProperties.swift
zihejia: it will be changed dynamically via our releasing script. There is a problem reading from Info.plist in SPM environment, the Info.plist can't be copied to the actually framework bundle.

PR #418, opened 2020-11-19
Fix events being dropped when flushing if there are corrupted items in the event queue

zihejia:
- Fix events being dropped when flushing if there are corrupted items in the event queue (filter out the bad event rather than dropping the whole batch of queue)
- safe guard the invalid numbers from being infinite or NaN before sending to Mixpanel

### DIFF
--- a/MixpanelDemo/MixpanelDemo/TrackingViewController.swift
+++ b/MixpanelDemo/MixpanelDemo/TrackingViewController.swift
@@ -22,6 +22,7 @@ class TrackingViewController: UIViewController, UITableViewDelegate, UITableView
                           "Register SuperProperties Once",
                           "Register SP Once w Default Value",
                           "Unregister SuperProperty"]
+    var eventNo: Int = 0
 
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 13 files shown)


MixpanelDemo/MixpanelDemo/TrackingViewController.swift
```diff
@@ -45,11 +46,12 @@ class TrackingViewController: UIViewController, UITableViewDelegate, UITableView
...
         switch indexPath.item {
         case 0:
-            let ev = "Track Event!"
+            eventNo += 1
+            let ev = "Track Event! No: \(eventNo)"
             Mixpanel.mainInstance().track(event: ev)
             descStr = "Event: \"\(ev)\""
         case 1:
-            let ev = "Track Event With Properties!"
+            let ev = "Track Event With Properties! No: \(eventNo)"
```
jaredmixpanel, line 54: in this example, I would make the event number a property and keep the event name consistent

zihejia: oops, thx for catching this

Sources/MixpanelInstance.swift
```diff
@@ -82,7 +82,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     open var trackAutomaticEventsEnabled: Bool? = nil
 
     /// Flush timer's interval.
-    /// Setting a flush interval of 0 will turn off the flush timer.
+    /// Setting a flush interval of 0 will turn off the flush timer and you need call flush API manually to upload queued data to the Mixpanel server.
```
jaredmixpanel, line 85:
```suggestion
    /// Setting a flush interval of 0 will turn off the flush timer and you need to call the flush() API manually to upload queued data to the Mixpanel server.
```

jaredmixpanel (approved):
lgtm

PR #425, opened 2020-12-30
Remove alias check

argenisf:
Removing a check for the alias within the identify function as it can prevent the ID from changing if alias has been called already.

### DIFF
--- a/Sources/MixpanelInstance.swift
+++ b/Sources/MixpanelInstance.swift
@@ -768,31 +768,31 @@ extension MixpanelInstance {
             }
 
 
-                if distinctId != self.distinctId {
-                    let oldDistinctId = self.distinctId
-                    self.alias = nil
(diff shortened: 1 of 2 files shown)


Sources/MixpanelInstance.swift
```diff
@@ -795,7 +793,7 @@ extension MixpanelInstance {
                 } else {
                     self.people.distinctId = nil
                 }
-            }
+
```
zihejia, line 796: minor nit: pls format the code. otherwise, it looks good.

argenisf: @zihejia I've fixed the formatting :)

argenisf: I've also added a fix to also encode "+" within URLs which was causing some decide requests not to fetch data.

PR #430, opened 2021-01-29
add clearTimedEvent(event: String) method

jaredmixpanel:


Sources/MixpanelInstance.swift
```diff
@@ -1414,6 +1414,17 @@ extension MixpanelInstance {
             }
         }
     }
+    
+    open func clearTimedEvent(event: String) {
```
zihejia, line 1418: could you add some doc?

jaredmixpanel: done!

zihejia (approved):
LGTM

Issue #431, opened 2021-02-09
2.7.7: Crash in Persistence.archiveToFile(_:object:token:)


igled7:
We are getting crashes in the method `Persistence.archiveToFile(_:object:token:)`

This started happening after upgrading to the version 2.7.7. Everything was working fine with 2.7.0.

**Environment**

- Xcode version: 11.7
- Mixpanel SDK version: 2.7.7 
- Installation method: Carthage

`EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x4d55545a0000040 Persistence.archiveToFile`

```
Crashed: com.mixpanel.archiveQueue
0  libswiftCore.dylib             0x19ff58b28 initializeWithCopy for Character + 20
1  libswiftCore.dylib             0x19fcd7d14 _SwiftDeferredNSDictionary.bridgeKeys() + 596
2  libswiftCore.dylib             0x19fcd8b94 _SwiftDeferredNSDictionary.getObjects(_:andKeys:count:) + 372
3  libswiftCore.dylib             0x19fce3a04 @objc _SwiftDeferredNSDictionary.getObjects(_:andKeys:count:) + 72
4  Foundation                     0x19d2806f4 -[NSDictionary(NSDictionary) encodeWithCoder:] + 244
5  Foundation                     0x19d3544f4 _encodeObject + 1212
6  Foundation                     0x19d25c534 -[NSKeyedArchiver _encodeArrayOfObjects:forKey:] + 380
7  Foundation                     0x19d2808bc -[NSDictionary(NSDictionary) encodeWithCoder:] + 700
8  Foundation                     0x19d3544f4 _encodeObject + 1212
9  Foundation                     0x19d2a3260 +[NSKeyedArchiver archiveRootObject:toFile:] + 236
10 Mixpanel                       0x1075a3dbc closure #1 in static Persistence.archiveToFile(_:object:token:) + 163 (Persistence.swift:163)
11 Mixpanel                       0x10755fc44 thunk for @escaping @callee_guaranteed () -> () + 28 (<compiler-generated>:28)
12 Mixpanel                       0x10754bae0 +[ExceptionWrapper try:catch:finally:] + 15 (ExceptionWrapper.m:15)
13 Mixpanel                       0x1075a7a3c specialized static Persistence.archiveToFile(_:object:token:) + 1012 (<compiler-generated>:1012)
[...]

PR #438, opened 2021-04-01
Fix crash in persistence.archiveToFile

zihejia:
Fix crash in persistence.archiveToFile caused by race conditions from variant changes and group updates. 
Address the issue: 
https://github.com/mixpanel/mixpanel-swift/issues/431


jaredmixpanel (approved):
lgtm

Issue #375, opened 2020-05-28
Support for Array<Optional<Double>>


heestand-xyz:
So the cast succeeds, tho I get a runtime crash...
```
let values: [Double?] = [nil, 0.0, 1.0]
if let mix: MixpanelType = values as? MixpanelType {
    print(mix)
}
```
> Conditional cast from '[Double?]' to 'MixpanelType' always succeeds

`Assertion failed: Property values must be of valid type (MixpanelType). Got Array<Optional<Double>>: file /Users/hexagons/App/Pods/Mixpanel-swift/Mixpanel/MixpanelType.swift, line 233`

jaredmixpanel:
@heestand-xyz we've added support for optional MixpanelType in v2.10.0

PR #447, opened 2021-05-15
Add support for optional MixpanelTypes

jaredmixpanel:
This PR addresses https://github.com/mixpanel/mixpanel-swift/issues/375 by adding general support for Optional `MixpanelType`'s


Sources/MixpanelType.swift
```diff
@@ -20,6 +20,52 @@ public protocol MixpanelType: Any {
...
+extension Optional: MixpanelType {
+    /**
+     Checks if this object has nested object types that Mixpanel supports.
+     Will always return true.
+     */
+    public func isValidNestedTypeAndValue() -> Bool {
+        if let val = self {
+            switch val {
+            case let mpt as MixpanelType:
+                return mpt.isValidNestedTypeAndValue()
```
zihejia, line 32: if it's MixpanelType, you could just return true

zihejia, line 32: looks good.

Sources/MixpanelType.swift
```diff
@@ -20,6 +20,52 @@ public protocol MixpanelType: Any {
     func equals(rhs: MixpanelType) -> Bool
 }
 
+extension Optional: MixpanelType {
+    /**
+     Checks if this object has nested object types that Mixpanel supports.
+     Will always return true.
+     */
+    public func isValidNestedTypeAndValue() -> Bool {
+        if let val = self {
```
zihejia, line 29: nit: 
use ```            guard let val = self else {
                return true
            }``` is prefered.

Sources/MixpanelType.swift
zihejia, line 32: nit: naming: can we choose a better name than `mpt`?

Sources/MixpanelType.swift
```diff
@@ -20,6 +20,52 @@ public protocol MixpanelType: Any {
...
+            return v == rhs as! Double
+        } else if let v = self as? Float, rhs is Float {
+            return v == rhs as! Float
+        } else if let v = self as? Bool, rhs is Bool {
+            return v == rhs as! Bool
+        } else if let v = self as? Date, rhs is Date {
+            return v == rhs as! Date
+        } else if let v = self as? URL, rhs is URL {
+            return v == rhs as! URL
+        } else if self is NSNull && rhs is NSNull {
```
zihejia, line 63: missing `nil`, `NSArray`, `Dictionary` and `Array`

zihejia: can we add some tests for `equals`?

zihejia (approved):
LGTM

zihejia (approved):

PR #496, opened 2022-01-10
fix  `First App Open` not always being able to be triggered

zihejia:
- Fix  `First App Open` not always being able to be triggered
- Add additional param `trackAutomaticEvents` in `initialize`: Whether or not to collect common mobile events

### DIFF
--- a/Sources/MixpanelInstance.swift
+++ b/Sources/MixpanelInstance.swift
@@ -214,7 +214,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     #endif // DECIDE
     
     #if !os(OSX) && !os(watchOS)
-    init(apiToken: String?, flushInterval: Double, name: String, optOutTrackingByDefault: Bool = false) {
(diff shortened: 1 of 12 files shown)


Sources/MixpanelInstance.swift
```diff
@@ -214,7 +214,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     #endif // DECIDE
     
     #if !os(OSX) && !os(watchOS)
-    init(apiToken: String?, flushInterval: Double, name: String, optOutTrackingByDefault: Bool = false) {
+    init(apiToken: String?, flushInterval: Double, name: String, optOutTrackingByDefault: Bool = false, trackAutomaticEvents: Bool = false) {
```
jaredmixpanel, line 217: default is `false` here? but `true` in Mixpanel.swift?

jaredmixpanel (approved):
lgtm, with just one question.

Sources/MixpanelInstance.swift
zihejia: should be `true`, good catch

PR #506, opened 2022-01-21
set content-type to application/json

jaredmixpanel:
set the content-type header to application/json to prevent certain payloads from being rejected by the server


jaredmixpanel:
test this please

zihejia (approved):
LGTM. don't forget to add tests later.

PR #508, opened 2022-01-24
fix build issue in JSONHandler

joetam:
3595189c6fec26ee86e92e8c96aa5ec09a23595f broke the build, by redeclaring the same variable `data` in the same scope


zihejia (approved):

zihejia:
thanks @joetam ,  thanks for bringing it up. It is working with Xcode 13 so we didn't catch it.

PR #511, opened 2022-01-31
Fix automatic events settings

zihejia:
`trackAutomaticEvents` param in `initialize`  should take precedence over `Autotrack` settings from the Mixpanel, but if `trackAutomaticEvents` is not set, should respect `Autotrack` settings from the Mixpanel.


jaredmixpanel (approved):
lgtm

PR #518, opened 2022-02-18
Allows using info.plist to override the appVersion used by Mixpanel

sidferreira:
Initial suggestions, pending related docs and tests


jaredmixpanel:
The `$app_version_string` property can be overridden by registering the desired value as a super property.

PR #519, opened 2022-02-19
Fix `disk I/O error` caused by race condition from multiple initializations with the same token

zihejia:


Sources/Mixpanel.swift
```diff
@@ -24,11 +24,11 @@ open class Mixpanel {
 
      - parameter token:                     your project token
      - parameter flushInterval:             Optional. Interval to run background flushing
-     - parameter instanceName:              Optional. The name you want to call this instance
+     - parameter instanceName:              Optional. The name you want to call this instance, has to be unique and correspond with the instance's project token
```
jaredmixpanel, line 27:
```suggestion
     - parameter instanceName:              Optional. The name you want to call this instance, must be unique 1:1 for each instance's project token. Defaults to project token.
```

jaredmixpanel (approved):
lgtm

gabriellanata:
This PR broke mixpanel on Mac. Specifically making the Mac initializer also use the `DispatchSemaphore` causes an app hang. Could we revert that to the previous behavior?

PR #521, opened 2022-03-03
check for ios app on mac in automatic props

jaredmixpanel:
This PR adds logic to check if the app is a "Designed for iPad" app running on macOS and if so, sets the `$os` to macOS. Unfortunately, there is no API that reports the correct macOS version for "Designed for iPad" apps running on macOS, so we omit it rather than mis-report


zihejia (approved):
LGTM

PR #523, opened 2022-03-04
Fix deadlock in initialization

gabriellanata:
Fixes https://github.com/mixpanel/mixpanel-swift/issues/524

I'm not exactly sure why this deadlocks in Mac when using SwiftUI (I'm on an M1). I've verified it only happens when initializing in a SwiftUI app. In any case, the semaphore should not be needed because it's the same behavior that `DispatchQueue.sync` does.

Changes in this PR:
1. Change `instanceQueue.async` to `instanceQueue.sync`.
2. Remove the any semaphore code.
3. Remove the `self` references to match code style now that the closure is non-escaping.

I've tested this works correctly and doesn't cause a deadlock on Mac M1. I have not tested how this affects iOS or other platforms.

If I may give an additional recommendation. I suggest changing the QoS of the `instanceQueue` dispatch queue from `.utility` to `.default`. This is a blocking sync operation that is usually executed at app initialization, running it at a `.utility` priority means it gets less system resources. Effectively it makes app initialization slower for any app that implements this SDK.

### DIFF
--- a/Sources/Mixpanel.swift
+++ b/Sources/Mixpanel.swift
@@ -159,15 +159,10 @@ class MixpanelManager {
                     trackAutomaticEvents: Bool? = nil,
                     useUniqueDistinctId: Bool = false,
                     superProperties: Properties? = nil) -> MixpanelInstance {
-        let semaphore = DispatchSemaphore(value: 0)
-        instanceQueue.async { [weak self] in
-            guard let self = self else {
-                return
-            }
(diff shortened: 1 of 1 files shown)


jaredmixpanel:
@gabriellanata many thanks for the detailed report, sample app and PR! Interestingly, this issue is not present on Intel Macs, definitely has something to do with the M1. We'll get your fixed merged in next week.

PR #525, opened 2022-03-05
Fix deadlock in initialization

jaredmixpanel:
Credit to @gabriellanata for https://github.com/mixpanel/mixpanel-swift/pull/523


zihejia (approved):
LGTM

PR #526, opened 2022-03-09
Check if automatic events flag is set before flushing

jaredmixpanel:
In the event that the `trackAutomaticEvents` flag has not been set, either in the initializer or by a /decide response, we will not flush any automatic event and instead hold them in the queue until one of the flags is set and upon next flush, we will discard them if the flag is false or flush them if the flag is true.


zihejia (approved):
LGTM

PR #542, opened 2022-05-04
pass completion handler to flush on background

jaredmixpanel:
This should hopefully fix: https://github.com/mixpanel/mixpanel-react-native/issues/132


zihejia (approved):
LGTM!

PR #543, opened 2022-05-06
get lib name and version from super props

jaredmixpanel:
pass lib name and version from super props into toggle sdk event


zihejia (approved):
LGTM

PR #545, opened 2022-06-22
Remove Decide and make trackAutomaticEvents required parameter

jaredmixpanel:


Mixpanel-swift.podspec
```diff
@@ -39,7 +39,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'Complete' do |ss|
     ss.ios.pod_target_xcconfig = {
-      'OTHER_SWIFT_FLAGS' => '$(inherited) -D DECIDE'
+      'OTHER_SWIFT_FLAGS' => '$(inherited) -D AUTOMATIC_EVENTS'
```
zihejia, line 42: can we remove this flag and only reply on the setting from `init`?

jaredmixpanel: replaced with `#if os(iOS) || os(tvOS)`

Sources/Track.swift
```diff
@@ -22,7 +22,10 @@ class Track {
     weak var mixpanelInstance: MixpanelInstance?
     
     var isAutomaticEventEnabled: Bool {
```
zihejia, line 24: nit: this function can be removed for simplicity.

jaredmixpanel: removed

PR #547, opened 2022-06-24
add an option for 'createAlias' for not calling identify

zihejia:
When you call the API `createAlias`, there is an implicit `identify` call inside the API done for you. This will keep your signup funnels working correctly in most cases. However, if that is not what you want, this PR will allow you to not call `identify` by specifying `andIdentify` to `false`.

 Please also note: With Mixpanel Identity Merge enabled, calling alias is no longer required but can be used to merge two IDs in scenarios where `identify` would fail

### DIFF
--- a/Sources/MixpanelInstance.swift
+++ b/Sources/MixpanelInstance.swift
@@ -612,6 +612,12 @@ extension MixpanelInstance {
                 self.hadPersistedDistinctId = true
             }
             
+            if self.userId == nil {
+                self.readWriteLock.write {
+                    self.userId = distinctId
+                }
+            }
+            
             if distinctId != self.distinctId {
                 let oldDistinctId = self.distinctId
                 self.readWriteLock.write {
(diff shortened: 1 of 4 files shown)


Sources/MixpanelInstance.swift
```diff
@@ -663,10 +666,12 @@ extension MixpanelInstance {
      - parameter alias:      A unique identifier that you want to use as an identifier for this user.
      - parameter distinctId: The current user identifier.
      - parameter usePeople: boolean that controls whether or not to set the people distinctId to the event distinctId.
+     - parameter andIdentify: an optional boolean that controls whether or not to call 'identify' with your current
+     user identifier(not alias). Default to true for keeping your signup funnels working correctly in most cases.
```
jaredmixpanel, line 670:
```suggestion
     user identifier (distinctId not alias). Defaults to true for keeping your signup funnels working correctly in most cases.
```

jaredmixpanel (approved):
LGTM

PR #549, opened 2022-07-02
add support for multiple instances under the same token

zihejia:
This PR will add support for multiple instances under the same token. Just specify an `instanceName` in `initialize`.
```
...

  /**
     Initializes an instance of the API with the given project token.
     Returns a new Mixpanel instance API object. This allows you to create more than one instance
     of the API object, which is convenient if you'd like to send data to more than
     one Mixpanel project from a single app.

     ...
     parameter instanceName:    Optional. The name you want to uniquely identify the Mixpanel Instance.
                                            It is useful when you want more than one Mixpanel instance under the same project token.
     ...
  */
    @discardableResult
    open class func initialize(token apiToken: String,
                               flushInterval: Double = 60,
                               instanceName: String? = nil,
                               optOutTrackingByDefault: Bool = false,
                               trackAutomaticEvents: Bool? = nil,
                               useUniqueDistinctId: Bool = false,
                               superProperties: Properties? = nil,
                               serverURL: String? = nil) -> MixpanelInstance {
```

### DIFF
--- a/MixpanelDemo/MixpanelDemoTests/MixpanelBaseTests.swift
+++ b/MixpanelDemo/MixpanelDemoTests/MixpanelBaseTests.swift
@@ -100,19 +100,19 @@ class MixpanelBaseTests: XCTestCase, MixpanelDelegate {
     }
     
     func eventQueue(token: String) -> Queue {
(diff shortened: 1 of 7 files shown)


jaredmixpanel (approved):
lgtm

Issue #550, opened 2022-07-14
Double value cannot be converted to Int because the result would be greater than Int.max (iOS 9.3.5 - Crash after updating from 3.1.0 to 3.5.0)


Pixette:
Hello!

Updating mixpanel pod from 3.1.0 to 3.5.0 causes my app to crash very frequently on iOS 9.3.5

Error is following:
![image](https://user-images.githubusercontent.com/83535288/179043899-5f7611ad-862c-46c6-8ba6-a9544194d72a.png)

Issue can always be reproduced on my iOS 9.3.5 device.
I reverted to 3.1.0 to solve the issue.

Thank you in advance for your help

repomies:
I have the same crash happen on Apple Watch, which apparently is using 32-bit integers (probably the same as @Pixette's app on an iOS 9 device).

For example, it's trying to use `epochInterval` value `1658180394`, which is multiplied by 1000 and then cast into an `Int`. The value of `Int.max` is `2147483647` on this device, so the assignment would be fine without the multiplication (until 19 January 2038, that is!), but with the multiplication it crashes big time.

repomies:
Here's the breaking commit from 3.3.0 (reverting to 3.2.6 fixed it for me): https://github.com/mixpanel/mixpanel-swift/commit/13c7cc96a4e490e8dac378639815d8ca86e623ad

jaredmixpanel:
@Pixette @repomies Sorry about that, thanks for the report. This should be fixed in 3.5.1. Feel free to reopen if the issue persists.

repomies:
@jaredmixpanel thanks for the very quick fix, seems to work for me!

PR #551, opened 2022-07-18
Send time as Decimal to avoid 32-bit max int

jaredmixpanel:
Fixes: https://github.com/mixpanel/mixpanel-swift/issues/550


zihejia (approved):
LGTM

PR #553, opened 2022-07-25
Mark final attribute in MixpanelManager

muukii:
it seems there is no subclass of MixpanelManager.
We could mark `final` attribute for it.
Which probably gains performance. Rather than that, readability will gain because we could notice this class won't have a subclass, so there is no need to be concerned about side effects in modifying functions and properties.


jaredmixpanel (approved):

jaredmixpanel:
thanks @muukii

PR #554, opened 2022-07-25
Use NSLock instead of DispatchQueue

muukii:
Thanks for this open-sourced library!

In the case of performing very small tasks, performing GCD takes a pretty large cost rather than actual operation inside the critical scope.
So I propose this PR that uses `NSLock` to get mutual-exclusion.

Referencing this note: https://gist.github.com/tclementdev/6af616354912b0347cdf6db159c37057
> queue.async() is wasteful if the dispatched block is small (< 1ms), as it will most likely require a new thread due to libdispatch's overcommit behavior. Prefer locking to protect shared state (rather than switching the execution context).

And I found some cases which are not protected from concurrent operations.

⚠️ This modification is still very part of this library. It affects consistency against the code entirely.

### DIFF
--- a/Sources/Mixpanel.swift
+++ b/Sources/Mixpanel.swift
@@ -210,32 +210,33 @@ class MixpanelManager {
     static let sharedInstance = MixpanelManager()
     private var instances: [String: MixpanelInstance]
     private var mainInstance: MixpanelInstance?
-    private let readWriteLock: ReadWriteLock
-    private let instanceQueue: DispatchQueue
+    
+    private let instancesLock = NSLock()
 
     init() {
         instances = [String: MixpanelInstance]()
         Logger.addLogging(PrintLogging())
-        readWriteLock = ReadWriteLock(label: "com.mixpanel.instance.manager.lock")
-        instanceQueue = DispatchQueue(label: "com.mixpanel.instance.manager.instance", qos: .utility)
     }
 
-    func initialize(token apiToken: String,
-                    flushInterval: Double,
-                    instanceName: String,
-                    optOutTrackingByDefault: Bool = false,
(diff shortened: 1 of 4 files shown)


Sources/Mixpanel.swift
```diff
@@ -210,52 +210,57 @@ class MixpanelManager {
     static let sharedInstance = MixpanelManager()
     private var instances: [String: MixpanelInstance]
     private var mainInstance: MixpanelInstance?
-    private let readWriteLock: ReadWriteLock
-    private let instanceQueue: DispatchQueue
```
muukii, line 214: It seems the separated lock and queue were not working well. 
there were some cases reading `mainInstance` inside readWriteLock's scope without mutex

Sources/Mixpanel.swift
```diff
@@ -264,30 +269,24 @@ class MixpanelManager {
...
+        instancesLock.with {
+            let instance = instances[instanceName]
+            
+            if instance == nil {
+                return
+            }
+            mainInstance = instance
         }
-        mainInstance = instance
     }
```
muukii, line 286: Might happen data-race in setMainInstance and getMainInstance

Sources/Mixpanel.swift
```diff
@@ -210,52 +210,57 @@ class MixpanelManager {
...
     func getInstance(name instanceName: String) -> MixpanelInstance? {
-        var instance: MixpanelInstance?
-        readWriteLock.read {
-            instance = instances[instanceName]
-        }
+        let instance: MixpanelInstance?
+        
+        instancesLock.lock()
+        instance = instances[instanceName]
+        instancesLock.unlock()
```
muukii, line 262: To use `let`, not using `NSLock.with`

Issue #564, opened 2022-10-26
Syntax error in "INSERT INTO" mixpanel events


markoengelman:
We are experiencing an issue since upgrading to version 4.0.X. (logs are from latest version 4.0.3)

Same as in this #556 , the tracking events are not showing up in Mixpanel Dashboard, and we are also getting an syntax error in the debug log from MixPanel. 

**Log**

```
2022-10-26 10:20:05.423072+0200 Mediately [34212:311314] [logging] near ".": syntax error in "INSERT INTO mixpanel_com.modrajagoda.mixpanel_events (data, flag, time) VALUES(?, ?, ?);"
2022-10-26 10:20:05.423915+0200 Mediately [34212:311314] [logging] near ".": syntax error in "CREATE TABLE IF NOT EXISTS mixpanel_com.modrajagoda.mixpanel_events(id integer primary key autoincrement,data blob,time real,flag integer);"
```

jaredmixpanel:
@markoengelman similar to #556 you're getting this error because of the `.` in your `instanceName` "com.modrajagoda.mixpanel", if you remove or replace the `.` in that string, for example "modrajagod" or "com_modrajagoda_mixpanel" this error should go away. However, in the next update I'll make sure we handle this so that special characters in `instanceName`'s don't cause this problem. Thanks for bringing it to our attention. I'll leave this issue open until that update is released.

jaredmixpanel:
@markoengelman this is fixed in v4.0.4

PR #566, opened 2022-10-26
Only use alphanumerics in MPDB token strings

jaredmixpanel:
Fix for https://github.com/mixpanel/mixpanel-swift/issues/564

### DIFF
--- a/Sources/MPDB.swift
+++ b/Sources/MPDB.swift
@@ -17,7 +17,7 @@ class MPDB {
     let apiToken: String
     
     init(token: String) {
-        apiToken = token.filter { !$0.isWhitespace }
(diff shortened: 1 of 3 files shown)


Sources/MPDB.swift
```diff
@@ -17,7 +17,7 @@ class MPDB {
     let apiToken: String
     
     init(token: String) {
-        apiToken = token.filter { !$0.isWhitespace }
```
zihejia, line 20: can we add a couple of tests for this?

jaredmixpanel: modified `testMPDB()` to use a token with white space and a bunch of special chars

zihejia (approved):
LGTM

Issue #565, opened 2022-10-26
Xcode 14 warnings


crayment:
Seeing these warnings when compiling our app on Xcode 14 with version 4.0.3

```
MixpanelInstance.swift#L615: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L707: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L790: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L821: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L896: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L988: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1043: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1065: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1141: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1158: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1168: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1183: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1197: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1208: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1227: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1255: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1284: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1315: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
MixpanelInstance.swift#L1329: Non-'@objc' instance method in extensions cannot be overridden; use 'public' instead
[...]

PR #568, opened 2022-11-02
Fix Xcode 14 warnings

jaredmixpanel:
Fixes https://github.com/mixpanel/mixpanel-swift/issues/565

image

### DIFF
--- a/Sources/AutomaticProperties.swift
+++ b/Sources/AutomaticProperties.swift
@@ -107,10 +107,7 @@ class AutomaticProperties {
                 }
             }
         }
-        if let model = String(validatingUTF8: modelCode) {
-            return model
-        }
-        return ""
+        return modelCode
     }
 
     #if os(watchOS)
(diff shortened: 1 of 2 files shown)


Sources/AutomaticProperties.swift
```diff
@@ -107,10 +107,7 @@ class AutomaticProperties {
                 }
             }
         }
-        if let model = String(validatingUTF8: modelCode) {
```
jaredmixpanel, line 110: `warning: 'init(validatingUTF8:)' is deprecated: Use a copy of the String argument` ... not sure exactly why we were doing this or if we really needed to? @zihejia

jaredmixpanel: https://developer.apple.com/documentation/swift/string/init(validatingutf8:)-2o7g5?changes=latest_major

zihejia (approved):
LGTM

PR #576, opened 2023-01-04
Updated ReadWrite Lock to enforce best practice

vikas171999:
- Corrected Spelling mistake
- Removed unneeded self
- made concurrentQueue private as it is not used outside that class. And as apple suggests if we are not using anything outside the class we should always make it private. We should upgrade the access level as and when the variable needed outside


Sources/ReadWriteLock.swift
```diff
@@ -8,19 +8,19 @@
...
+    private let concurrentQueue: DispatchQueue
 
     init(label: String) {
-        self.concurentQueue = DispatchQueue(label: label, qos: .utility, attributes: .concurrent)
+        concurrentQueue = DispatchQueue(label: label, qos: .utility, attributes: .concurrent)
     }
 
     func read(closure: () -> Void) {
-        self.concurentQueue.sync {
+        concurrenttQueue.sync {
```
jaredmixpanel, line 18:
```suggestion
        concurrentQueue.sync {
```

jaredmixpanel (changes requested):
small typo and please also pull in the latest changes from master, to get updated CI config.

Sources/ReadWriteLock.swift
vikas171999: Done

jaredmixpanel:
@vikas171999 CI is still failing. Please merge the latest upstream master to get the updates to our CI config.

jaredmixpanel:
@vikas171999 can you please pull in the latest changes from master? they include some CI config changes that will fix your CI failure.

PR #581, opened 2023-02-03
SwiftUI preview fix

RamblinWreck77:
Don't fire this assert in the simulator (SwiftUI Preview), as the new SwiftUI lifecycle doesn't initialize Main view controller in child views so the SDK will likely not be configured


RamblinWreck77:
@jaredmixpanel Done! Looks like that worked, thanks

jaredmixpanel (approved):
thank you!

Issue #582, opened 2023-02-06
Can get into a state where flush timer is scheduled with 0 timeInterval


julsh:
We observed some odd behavior with our freshly set up MixPanel integration where we saw in-app memory grow steadily while the app was sitting idle. Some investigation brought up the following:

```
// initialize mixpanel with our token
Mixpanel.initialize(token: mixpanelProjectToken, trackAutomaticEvents: true)
// manually control flush
Mixpanel.mainInstance().flushInterval = 0
```

We want to manually control the flushing and thus set the `flushInterval` to 0, according to the documentation this is how to do that. However, due to the fact that the scheduling of the flush timer inside `Flush.swift` is done async on the main thread, the following order of events occurs:
- MixPanel gets initialized with the default `flushInterval` of 60
- `startFlushTimer()` gets called, checks `if flushInterval > 0` (which is true at this point), schedules the block that initializes the timer on the main thread
- We manually set `Mixpanel.mainInstance().flushInterval = 0` 
- The main thread executes the block, and without an additional check for `flushInterval > 0`, the timer will now be scheduled with a 0 interval

I've since realized that `MixPanel.initialize(...)` can (and probably should be!) be called with the custom `flushInterval` parameter, which would avoid getting into this state. But you'll probably want to tighten up this "loophole" anyway, since it doesn't necessarily seem like an incorrect setup to initialize with default params and then do the custom setup in a next step. I'll open a PR with a fix for this.

jaredmixpanel:
@julsh your PR is merged and released, closing this for now, feel free to re-open if the issue persists.

PR #583, opened 2023-02-06
Check flush interval > 0 inside the dispatch block

julsh:
The value could have been modified between evaluating the statement and executing the block.

See https://github.com/mixpanel/mixpanel-swift/issues/582


jaredmixpanel:
@julsh can you please pull in the latest changes from master? they include some CI config changes that will fix your CI failure.

PR #588, opened 2023-02-23
Access the timedEvents property inside of a readWriteLock.

prydain007:
Access the timedEvents property inside of a readWriteLock.  Fixes crashes when client tries to read and write to timedEvents property at the same time.


jaredmixpanel:
@prydain007 can you please pull in the latest changes from master? they include some CI config changes will fix your CI failure.

Issue #594, opened 2023-03-16
Flush grows memory significantly until crash


thibauddavid:
Hi,

I'm using MixPanel 4.0.6.

When my share extension (which has a memory size limit) tries to flush, the memory grows until crashing.

It seems like `loadEntitiesInBatch` always uses a batchSize of Int.max, leading to `readRows` not using any limit for the SELECT query.
As a result, `rows.append(entity)` makes `rows` array growing unbounded.

I think `loadEntitiesInBatch` should be called in a paginated way for flushing to be memory-consistent.

zihejia:
hi @thibauddavid , thanks for bringing it up. We will fix it.

zihejia:
hi @thibauddavid, the fix is in 4.1.0. I'm closing this one now, feel free to reopen it if the problem exists. Thanks again!

PR #596, opened 2023-03-22
Set the number of events per flush to 1,000 to reduce memory footprint

zihejia:
This will greatly improve the performance and reduce crashes. Thanks @thibauddavid, #594

### DIFF
--- a/Sources/MixpanelInstance.swift
+++ b/Sources/MixpanelInstance.swift
@@ -46,6 +46,7 @@ protocol AppLifecycle {
 
 /// The class that represents the Mixpanel Instance
 open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDelegate {
+    
     /// apiToken string that identifies the project to track data to
     open var apiToken = ""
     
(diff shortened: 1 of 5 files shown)


Sources/MixpanelInstance.swift
```diff
@@ -925,11 +925,11 @@ extension MixpanelInstance {
...
             // automatic events will NOT be flushed until one of the flags is non-nil
             let eventQueue = self.mixpanelPersistence.loadEntitiesInBatch(
-                type: self.persistenceTypeFromFlushType(.events),
+                type: self.persistenceTypeFromFlushType(.events), batchSize: APIConstants.flushSize,
                 excludeAutomaticEvents: !self.trackAutomaticEventsEnabled
             )
-            let peopleQueue = self.mixpanelPersistence.loadEntitiesInBatch(type: self.persistenceTypeFromFlushType(.people))
-            let groupsQueue = self.mixpanelPersistence.loadEntitiesInBatch(type: self.persistenceTypeFromFlushType(.groups))
+            let peopleQueue = self.mixpanelPersistence.loadEntitiesInBatch(type: self.persistenceTypeFromFlushType(.people), batchSize: APIConstants.flushSize)
+            let groupsQueue = self.mixpanelPersistence.loadEntitiesInBatch(type: self.persistenceTypeFromFlushType(.groups), batchSize: APIConstants.flushSize)
```
jaredmixpanel, line 932: what if there are more than 1000 events? should we automatically queue the next flush call? we don't want to knowingly leave events un-flushed while we have the opportunity to flush them. it might be more than 5 days until the next flush and the events will be rejected.

jaredmixpanel (approved):
LGTM

Issue #578, opened 2023-01-12
Lots of crashes on `FlushRequest.sendRequest(_:type:useIP:completion:)` with 4.0.4


malcommac:
**Integration Method:** SPM
**Xcode Version:** 14.1.0
**Library Version:** 4.0.4
**Platform:** iOS
**Language:** Swift
**Description:** 

We got lots of crashes using the 4.0.4 release on `FlushRequest.sendRequest(_:type:useIP:completion:)`:

<img width="739" alt="image" src="https://user-images.githubusercontent.com/235645/208411373-150746f0-0c1a-4364-b921-ca2f9b7d3110.png">

The raw stack trace:

- From Crashlytics: [1000_issue_bdea1234c542cc5311cd883d0eaa07fe_crash_session_7f2766abffd840518ee45bbfe240ac1f_DNE_0_v2_stacktrace.txt](https://github.com/mixpanel/mixpanel-iphone/files/10258669/1000_issue_bdea1234c542cc5311cd883d0eaa07fe_crash_session_7f2766abffd840518ee45bbfe240ac1f_DNE_0_v2_stacktrace.txt)
- From Sentry: [938f83131e154ff9b6d9fcbacb151413-symbolicated.txt](https://github.com/mixpanel/mixpanel-swift/files/10400468/938f83131e154ff9b6d9fcbacb151413-symbolicated.txt)

malcommac:
anyone working on this SDK anymore?

jaredmixpanel:
@malcommac Are you able to reproduce this issue? If so, can you provide the steps? Were you using older versions without this problem? If so, what version were you upgrading from? Any additional info you can provide would be helpful.

malcommac:
Before that, we used 3.5.1 without the issue. Unfortunately, I'm not able to reproduce this issue, but it seems to be happening at the early stage of the app launch. Maybe it could be related to some flush events of the local events not sent in the previous session?

malcommac:
I wonder if anyone read these issues. Anyway it still the most relevant crash issue in our app.


[...]

PR #611, opened 2023-07-18
Re-work thread safety mechanisms for flush process

jaredmixpanel:
Hopefully a fix for https://github.com/mixpanel/mixpanel-swift/issues/578


Sources/Flush.swift
```diff
@@ -86,53 +89,47 @@ class Flush: AppLifecycle {
...
-                (entity["id"] as? Int32) ?? 0
-            }
-            // Log data payload sent
-            Logger.debug(message: "Sending batch of data")
-            Logger.debug(message: batch as Any)
-            let requestData = JSONHandler.encodeAPIData(batch)
-            if let requestData = requestData {
-                let semaphore = DispatchSemaphore(value: 0)
-                #if os(iOS)
+        flushBatchQueue.sync {
```
jaredmixpanel, line 92: put all calls to `flushQueueInBatches` into sync queue in case it's being called from multiple threads

Sources/Flush.swift
```diff
@@ -86,53 +89,47 @@ class Flush: AppLifecycle {
...
-                                            semaphore.signal()
-                })
-                _ = semaphore.wait(timeout: DispatchTime.distantFuture)
-            }
-            if !shouldContinue {
-                break
+                    #endif // os(iOS)
+                    let success = flushRequest.sendRequest(requestData,
+                                                           type: type,
+                                                           useIP: useIPAddressForGeoLocation)
```
jaredmixpanel, line 113: move the semaphore into `sendRequest` and make it synchronous from the perspective of `flushQueueInBatches`

Sources/FlushRequest.swift
```diff
@@ -39,14 +38,19 @@ class FlushRequest: Network {
...
+        var result = false
+        let semaphore = DispatchSemaphore(value: 0)
         flushRequestHandler(BasePath.getServerURL(identifier: basePathIdentifier),
                             resource: resource,
                             completion: { success in
-                                completion(success)
+                                result = success
+                                semaphore.signal()
         })
+        _ = semaphore.wait(timeout: DispatchTime.distantFuture)
```
jaredmixpanel, line 49: Maybe, we should change this timeout to some number of minutes?

zihejia: yeah, agreed.

Sources/Flush.swift
zihejia, line 92: calling .sync could potentially cause the deadlock.

zihejia (approved):
LGTM

RamblinWreck77:
Sorry to comment on an already-merged MR, but are you guys open to a Swift Concurrency/async-await refactor?

I have a bunch of concerns with Flush.swift:

- flushIntervalReadWriteLock is on a .utility QoS _concurrent_ queue

- DispatchQueue.main.async is forcing work to the main thread, when mixpanel IMO should pretty much never touch the main thread

- I suspect there's a priority inversion there locking with a lower QoS and doing work on main

- calling out to objc #selectors inside a DispatchQueue.main block that weakly calls a delegate function sets off my spidey-sense

It seems like a lot of what you want to accomplish could be better handled in pure-swift with:

- Swift Concurrency Tasks

- async-await

- Swift Actors for mutable state isolation

@zihejia @jaredmixpanel I can see you guys support back to iOS 11 (!!) in the Package.swift

Would you consider raising the minimum OS to iOS 13.0, or even 14.0? Apple back-ported Swift Concurrency to 13.0, and that unlocks a whole world of feature improvements.

If some PM gripes about legacy users, you could always fork this repo (as it is fairly stable) as mixpanel_legacy, and leave it up who insist on using Paleolithic-era OSes while the main repo forges ahead.

PR #617, opened 2023-11-10
add a new property 'flushBatchSize' for fine tuning the network request

zihejia:
Add the following property `flushBatchSize`:

The flushBatchSize property controls the number of events sent in a single network request to the Mixpanel server.
Batch size affects how data is flushed from the client to Mixpanel. By configuring the size of the batch, you can
optimize network usage and control how frequently the client communicates with the server.

Usage: 

`mixpanel.flushBatchSize = 20`

### DIFF
--- a/Sources/MixpanelInstance.swift
+++ b/Sources/MixpanelInstance.swift
@@ -119,15 +119,15 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
         }
     }
     
-    /// The flushBatchSize property controls the number of events sent in a single network request to the Mixpanel server.
-    /// Batch size affects how data is flushed from the client to Mixpanel. By configuring the size of the batch, you can
-    /// optimize network usage and control how frequently the client communicates with the server.
+    /// The `flushBatchSize` property determines the number of events sent in a single network request to the Mixpanel server.
+    /// By configuring this value, you can optimize network usage and manage the frequency of communication between the client
+    /// and the server. The maximum size is 50; any value over 50 will default to 50.
     open var flushBatchSize: Int {
         get {
             return flushInstance.flushBatchSize
         }
         set {
-            flushInstance.flushBatchSize = newValue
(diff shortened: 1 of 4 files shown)


Sources/MixpanelInstance.swift
```diff
@@ -119,6 +119,19 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
         }
     }
     
+    /// The flushBatchSize property controls the number of events sent in a single network request to the Mixpanel server.
+    /// Batch size affects how data is flushed from the client to Mixpanel. By configuring the size of the batch, you can
+    /// optimize network usage and control how frequently the client communicates with the server.
```
jaredmixpanel, line 124: `/track` has a maximum batch size of 50: https://developer.mixpanel.com/reference/track-event#limits

zihejia: good catch!

jaredmixpanel (approved):
LGTM but we should enforce the max batch size of 50 for the `/track` endpoint

RamblinWreck77:
@jaredmixpanel Sorry to necro a merged MR, but am I reading this correctly?

Screenshot 2023-11-13 at 5 32 17 PM

It appears to me that a call to `flush(performFullFlush: true)` would attempt an upload with Int.max batch size, bypassing the  flushBatchSize limit imposed by the API? What does the API do if you send it more than 50? Drop events?

zihejia:
> @jaredmixpanel Sorry to necro a merged MR, but am I reading this correctly?
> 
> Screenshot 2023-11-13 at 5 32 17 PM
> 
> It appears to me that a call to `flush(performFullFlush: true)` would attempt an upload with Int.max batch size, bypassing the flushBatchSize limit imposed by the API? What does the API do if you send it more than 50? Drop events?
> 
> I guess it makes sense that a "full" flush will send everything, but I don't think anyone would want that if it resulted in lost events.

Hi @RamblinWreck77, if you set the batch size to more than 50, it won't drop events but just fall back to 50. perhaps we should make it clearer in the doc.

Issue #612, opened 2023-08-09
Support visionOS


frlefebvre:
My current app currently does not compile because of the Mixpanel library. As far as I can tell, all it needs is a few `#if os(visionOS)` statements to be added so the right libraries get imported.

Keep up the good work!

jaredmixpanel:
@frlefebvre Please feel free to submit a PR and we'll take a look.

PR #618, opened 2023-11-10
Fixes so project builds for visionOS

JagCesar:
Fixes so Mixpanel builds in both Xcode 15 prod and betas, making it possible to work with visionOS targets.

fixes mixpanel/mixpanel-swift#612


zihejia (approved):
thank you so much @JagCesar !

Issue #614, opened 2023-09-27
How to track multiple "time" events with the same name


fzy-spyro:
Hey Mixpanel team, 

I have a question - is it possible to track multiple "time" events withe the same name? 

Let's imagine that I am working on a "cloud storage" app and I want to track how long it takes for a user to upload a single file to a server and at the same time app allows uploading multiple files in parallel. 

my tracking code would look smth like: 

```swift

func startUpload(filePaths: [String]) {
 filePaths.forEach { upload($0) }
}

func upload(filePath: String) async {
  Task {
    do {
      Mixpanel.mainInstance().time("FileUpload")
      try await httpClient.uploadFile(filePath)
      Mixpanel.mainInstance().track("FileUpload", properties: ["result": "ok", "path": filePath])
    } catch {
      Mixpanel.mainInstance().track("FileUpload", properties: ["result": "fail", "path": filePath])      
    }
  }
}
```

Is this valid code from Mixpanel tracking point of view? Will it correctly track eg. 3 uploads happening in parallel? 

If not how this can be achieved using mixpanel?

nathanmclain:
@fzy-spyro ever get an answer or find a work around on this?

fzy-spyro:
@nathanmclain unfortunately no. I implemented my own way of tracking those events

cmriboldi:
@nathanmclain and @fzy-spyro I fixed the issue in [this MR](https://github.com/mixpanel/mixpanel-swift/pull/625). It would be nice if we get some eyes and traction on it so that we can get this merged in so that we don't have to make our own "fixes" for this very fixable issue with the SDK.

PR #625, opened 2024-01-22
Cmriboldi/add event i ds

cmriboldi:
As mentioned in [this issue](https://github.com/mixpanel/mixpanel-swift/issues/614) there is no way to time events independently of their name. This is a HUGE issue for any serious user of Mixpanel events because there is no way to follow the [Best Practice Naming Guidelines](https://docs.mixpanel.com/docs/data-structure/events-and-properties#best-practices) and be certain your durations are correct.

For example, if I were to have a generic System event that I want to time as long as I have one of those events I'm fine, but if I ever need to time multiple system events there is no way for me to organize them or distinguish using Mixpanel which `.track(event:)` call matches with the original `time(event:)` call. This means that durations will be completely missing or inaccurate.

The that I'm adding in this MR is to link the tracking of timed events to an EventID instead of to the event name. That means callers of the SDK can provide their own unique id's for id's of events and be guaranteed that when the duration completes for that generic event the duration is completely accurate.

I added a new function signature for these timed and tracked events that allow users to provide a String eventID to link the two events. I would have liked to create a more type safe ID so that it's not just any string but because of the way the SDK works with events I didn't want to change too much besides separating the idea of an eventName with the idea of an eventID. I also separated out the idea of a `TimedEvents` dictionary from the `InternalProperties` dictionary that existed previously, this adds some small type safety around the dictionary and adds purpose clarity in the code.

The original method signatures still exist and work just as they did before they simply call the eventID versions of these methods so that there aren't two places that the logic needs to be maintained.

### DIFF
--- a/Sources/MixpanelInstance.swift
+++ b/Sources/MixpanelInstance.swift
@@ -37,6 +37,8 @@ public protocol MixpanelDelegate: AnyObject {
 
 public typealias Properties = [String: MixpanelType]
 typealias InternalProperties = [String: Any]
+typealias EventID = String
+typealias TimedEvents = [EventID: TimeInterval]
 typealias Queue = [InternalProperties]
 
 protocol AppLifecycle {
@@ -210,7 +212,7 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
     var networkQueue: DispatchQueue
     var optOutStatus: Bool?
     var useUniqueDistinctId: Bool
-    var timedEvents = InternalProperties()
+    var timedEvents = TimedEvents()
     
     let readWriteLock: ReadWriteLock
 #if os(iOS) && !targetEnvironment(macCatalyst)
@@ -819,7 +821,7 @@ extension MixpanelInstance {
             
             MixpanelPersistence.deleteMPUserDefaultsData(instanceName: self.name)
             self.readWriteLock.write {
-                self.timedEvents = InternalProperties()
+                self.timedEvents = TimedEvents()
                 self.anonymousId = self.defaultDeviceId()
                 self.distinctId = self.addPrefixToDeviceId(deviceId: self.anonymousId)
                 self.hadPersistedDistinctId = true
@@ -1027,7 +1029,7 @@ extension MixpanelInstance {
             guard let self = self else {
                 return
             }
-            var shadowTimedEvents = InternalProperties()
+            var shadowTimedEvents = TimedEvents()
             var shadowSuperProperties = InternalProperties()
             
             self.readWriteLock.read {
@@ -1187,12 +1189,12 @@ extension MixpanelInstance {
      - parameter event: the name of the event to be tracked that was passed to time(event:)
      */
     public func eventElapsedTime(event: String) -> Double {
-        var timedEvents = InternalProperties()
+        var timedEvents = TimedEvents()
         self.readWriteLock.read {
             timedEvents = self.timedEvents
         }
         
-        if let startTime = timedEvents[event] as? TimeInterval {
+        if let startTime = timedEvents[event] {
             return Date().timeIntervalSince1970 - startTime
         }
         return 0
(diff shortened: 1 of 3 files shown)


Sources/MixpanelInstance.swift
```diff
@@ -37,6 +37,8 @@ public protocol MixpanelDelegate: AnyObject {
 
 public typealias Properties = [String: MixpanelType]
 typealias InternalProperties = [String: Any]
+typealias TimedEventID = String
+typealias TimedEvents = [TimedEventID: TimeInterval]
```
cmriboldi, line 41: I added these two typealiases for additional type-safely and clarity. Ideally TimedEventID is it's own type that wraps a string but I didn't want to change the call site too drastically from what it already is.

Sources/MixpanelInstance.swift
```diff
@@ -1017,17 +1020,34 @@ extension MixpanelInstance {
      - parameter properties: properties dictionary
      */
     public func track(event: String?, properties: Properties? = nil) {
+        track(event: event, withID: nil, properties: properties)
```
cmriboldi, line 1023: The call site for all of these methods remains the same for backwards compatibility, and under the hood they just call the other method that uses the event id.

Sources/Track.swift
```diff
@@ -32,28 +32,30 @@ class Track {
...
+        if let event {
+            eventName = event
         } else {
             Logger.info(message: "mixpanel track called with empty event parameter. using 'mp_event'")
         }
-        if !(mixpanelInstance?.trackAutomaticEventsEnabled ?? false) && ev.hasPrefix("$ae_") {
+        if !(mixpanelInstance?.trackAutomaticEventsEnabled ?? false) && eventName.hasPrefix("$ae_") {
             return timedEvents
         }
+        let timedEventID = eventID ?? eventName
```
cmriboldi, line 50: If no eventID is provided we fall back to using whatever the eventName is which means we default back to the original behavior.

Sources/Track.swift
```diff
@@ -139,32 +141,32 @@ class Track {
         update(&superProperties)
     }
 
-    func time(event: String?, timedEvents: InternalProperties, startTime: Double) -> InternalProperties {
+    func time(eventID: TimedEventID, timedEvents: TimedEvents, startTime: TimeInterval) -> TimedEvents {
```
cmriboldi, line 144: I renamed the property name for clarity, changed the types for consistency/correctness and removed the optionality of the eventID because it didn't actually need to be optional.

cmriboldi:
@zihejia and @jaredmixpanel Is there any chance we can add this? I can go ahead and fix the merge conflict but this is a great enhancement we would love to get into your SDK so that we don't have to live off of a fork of the repo.

jaredmixpanel:
@cmriboldi Thanks for the PR. I think we can add it, just a few things though...

1. Fix the conflict
2. Add/update tests

- https://github.com/mixpanel/mixpanel-swift/blob/master/MixpanelDemo/MixpanelDemoTests/MixpanelDemoTests.swift#L787
- https://github.com/mixpanel/mixpanel-swift/blob/master/MixpanelDemo/MixpanelDemoTests/MixpanelOptOutTests.swift#L208
- https://github.com/mixpanel/mixpanel-swift/blob/master/MixpanelDemo/MixpanelDemoMacTests/MixpanelDemoTests.swift#L714
- https://github.com/mixpanel/mixpanel-swift/blob/master/MixpanelDemo/MixpanelDemoMacTests/MixpanelOptOutTests.swift#L214

3. Instead of labeling the new parameter `eventID`, I'd suggest to be clearer/more explicit and call it: `timedEventID` 
4. Maybe make the type [UUID](https://developer.apple.com/documentation/foundation/uuid)?

cmriboldi:
Awesome! thank you @jaredmixpanel. All of those are reasonable. I'll make those changes and then update this MR.

cmriboldi:
@jaredmixpanel Sorry it took so long but here's the fixed version of this PR. I resolved the conflicts, incorporated your suggested changes, and added tests as well. Some of the unrelated tests were stale but I fixed some of them.

PR #626, opened 2024-01-23
Expose publicly some types and methods related to the logging system

jduquennoy:
The goal of this PR is to allow users to inject custom loggers to Mixpanel, so that mixpanel's logs can be send to the same destination used by the host application.


jaredmixpanel:
@jduquennoy Thanks for the PR. Since the types were being made public, I prefixed them with `Mixpanel`. I will make a new release with this included soon.

jaredmixpanel:
@jduquennoy This has been released with `v4.3.1`

PR #627, opened 2024-01-30
visionOS Operating System & Version along with Automatic Events

ay8s:
This adds support for sending os and os_version for visionOS along with Automatic Events.

I'm sure there is more parts of mixpanel-swift that would function on visionOS that I've missed.


zihejia (approved):
thank you so much @ay8s

PR #628, opened 2024-02-09
feat: Proxy server header capability

mohitanand-cred:
Changes:

 - Added support for custom `header` and `query items` parameter.
 - New struct called `ProxyServerConfig` which takes in proxy server url and a delegate(`MixpanelProxyServerDelegate`).
 - `MixpanelProxyServerDelegate` can be used to provide headers or query items to the proxy server. The delegate func `mixpanelResourceForProxyServer(_ mixpanel: MixpanelInstance) -> ServerProxyResource?` is being called when its time flush the batch request.
  - Created a new `initialize` function in Mixpanel which takes in `ProxyServerConfig` instead of just `serverURL`.


Sources/Mixpanel.swift
```diff
@@ -56,6 +56,49 @@ open class Mixpanel {
...
+    
+    @discardableResult
+    open class func initialize(token apiToken: String,
+                               trackAutomaticEvents: Bool,
+                               flushInterval: Double = 60,
+                               instanceName: String? = nil,
+                               optOutTrackingByDefault: Bool = false,
+                               useUniqueDistinctId: Bool = false,
+                               superProperties: Properties? = nil,
+                               proxyServerConfig: ProxyServerConfig) -> MixpanelInstance {
```
zihejia, line 92: if `proxyServerConfig` is optional, could you add `proxyServerConfig: ProxyServerConfig? = nil` to the existing `initialize` function rather than adding a new function?

zihejia:
thank you so much @mohitanand-cred ! the PR looks good except one comment, also could you rebase/merge from the latest master and make sure all CI tests pass?

Sources/Mixpanel.swift
mohitanand-cred: Thanks for the feedback on the PR 🙌 .  On using the same `intialize` function, they are few points I want to highlight.

- By combining `serverURL` and `ProxyServerConfig` into one method, we risk users setting both parameters, which doesn't align with their intended use. Ideally one of them should be set. Keeping them separate ensures users have a clear choice without the possibility of configuring conflicting options.
- With two distinct initialization methods, we're directly guiding users on their options. **One method is for basic server URL configuration, and the other is for more advanced proxy configurations.**

jaredmixpanel (approved):
LGTM

Issue #620, opened 2023-11-28
Support Privacy Manifest


RyotaIwai:
Please let me know when Privacy Manifest will be supported.
Apple has declared that this support will become required in the spring of 2024.

Thank you in advance.

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files

RyotaIwai:
Our application needs to be supported by May 1, 2024.
What is the timeline for Mixpanel’s Privacy Manifest support?

https://developer.apple.com/news/?id=3d8a9yyh

xyn1nja:
Hey @RyotaIwai as I understand Mixpanel is actually not included in the [list of SDKs that require a privacy manifest](https://developer.apple.com/support/third-party-SDK-requirements/) so it's entirely optional and eng will not be prioritising it. If you really want to include it for your app, you can refer to this: https://mixpanel.com/legal/app-store-privacy-details/

RyotaIwai:
@xyn1nja 

Hi.
I know Mixpanel is not on Apple's list.
But I understand that by May 1 our App needs to support Privacy Manifest.

For that we need to know about the API and data handling that Mixpanel is dealing with.
Is there no plan for Mixpanel to support Privacy Manifest at this time?

nhelke:
@xyn1nja while I understand that Mixpanel is not singled out on Apple's list (yet), my reading of the [new requirements](https://developer.apple.com/news/?id=3d8a9yyh) vis-a-vis "need[ing] to declare approved reasons for using [[required reason APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)]" applies to app submissions as a whole, meaning that if Mixpanel uses any of those APIs and does not provide a privacy manifest, the app submission will be rejected.

I guess we will find out on March 13 if we start getting warnings when submitting.

benMohamed:
You need to update your code to support Apple's privacy requirements; otherwise, we will be forced to change your SDK :(

benMohamed:
Me too I got this message because of Mixpanel sdk
`
 
--
 
[...]

PR #633, opened 2024-03-14
Add privacy manifest

jaredmixpanel:
This PR adds a privacy manifest to declare the use of the required API for UserDefaults with the following reason:

```
1C8F.1
Declare this reason to access user defaults to read and write information that is only accessible to the apps, app extensions, and App Clips that are members of the same App Group as the app itself.

This reason does not permit reading information that was written by apps, app extensions, or App Clips outside the same App Group or by the system. This reason also does not permit writing information that can be accessed by apps, app extensions, or App Clips outside the same App Group.
```

Addresses: https://github.com/mixpanel/mixpanel-swift/issues/620

### DIFF
--- a/Mixpanel.xcodeproj/project.pbxproj
+++ b/Mixpanel.xcodeproj/project.pbxproj
@@ -95,6 +95,7 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXFileReference section */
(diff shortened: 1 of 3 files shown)


zihejia (approved):
LGTM

PR #700, opened 2026-03-17
Add loadFlags(completion:) callback variant

msiebert:
## Summary
- Adds `loadFlags(completion: ((Bool) -> Void)?)` to the `MixpanelFlags` protocol, allowing callers to receive a success/failure callback when flags finish loading.
- Implements the new method in `FeatureFlagManager`, delegating to `_fetchFlagsIfNeeded(completion:)`.
- Refactors the existing no-arg `loadFlags()` to delegate to `loadFlags(completion: nil)`.

## Test plan
- [ ] Verify `loadFlags()` (no-arg) still works as before
- [ ] Verify `loadFlags(completion:)` calls back with `true` on successful flag fetch
- [ ] Verify `loadFlags(completion:)` calls back with `false` on failure (e.g., network error)
- [ ] Confirm protocol conformance compiles for any existing adopters of `MixpanelFlags`

### DIFF
--- a/MixpanelDemo/MixpanelDemoTests/MixpanelDemoTests.swift
+++ b/MixpanelDemo/MixpanelDemoTests/MixpanelDemoTests.swift
@@ -218,6 +218,11 @@ class MixpanelDemoTests: MixpanelBaseTests {
       loadFlagsCallCount += 1
     }
 
+    func loadFlags(completion: ((Bool) -> Void)?) {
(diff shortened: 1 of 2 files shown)


ketanmixpanel (approved):
Looks good.

quintonpryce:
A nice enhancement would be to pass the error back or add a throw. cc @msiebert

PR #705, opened 2026-03-31
Use iOS file-protection flag when opening DB

santigracia:
Add SQLITE_OPEN_FILEPROTECTION_COMPLETEUNTILFIRSTUSERAUTHENTICATION (0x00300000) to sqlite3_open_v2 flags on iOS so the database and its WAL/SHM files remain accessible in the background after first unlock. This prevents WAL-related fsync/pwrite hangs (and potential watchdog kills) when the device is locked during background execution. Uses conditional compilation to apply the flag only on iOS; non-iOS platforms (macOS, wachOS, etc) keep the existing read/write/create flags as they don't have the same locked-screen file protection behavior that causes the hanging I/O on iOS

### DIFF
--- a/Sources/MPDB.swift
+++ b/Sources/MPDB.swift
@@ -55,7 +55,18 @@ class MPDB {
       return
     }
     if let dbPath = pathToDb() {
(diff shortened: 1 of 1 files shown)


ketanmixpanel (approved):
Looks good.

PR #715, opened 2026-04-16
Fix feature flags and track events race condition

ketanmixpanel:
This pull request improves the way feature flag operations are synchronized and documents important threading considerations for synchronous flag retrieval methods. The main changes are the introduction of a configurable `trackingQueue` for all flag operations, enhanced documentation and runtime warnings to discourage blocking calls on the main thread, and refactoring to ensure all relevant operations use the correct queue.

**Important**

- Main thread blocking - If user calls getVariantSync() or getVariantValueSync() from main thread after track(), it will block while waiting for variant activation. This is by design to fix the race, but could cause UI hiccups if the async work takes long. Our warnings/DEBUG checks address this.                                                                                                                                                                                                   

**Threading and Synchronization Improvements:**

* Introduced a `trackingQueue` property to `FeatureFlagManager`, allowing feature flag operations to be dispatched on a configurable queue, with a fallback to a global queue if not provided. Added a convenience initializer for dependency injection of the queue. (`Sources/FeatureFlags.swift`, [[1]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R348-R350) [[2]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R363-R373) [[3]](diffhunk://#diff-764368484691977ff02e15390a9fac71fa56690bb43a46b1d99ab3e5d6dd50c5L394-R394)
* Refactored all asynchronous flag operations (such as `loadFlags`, `getAllVariants`, and first-time event checks) to use `getTrackingQueue()` instead of hardcoding `DispatchQueue.global(qos: .userInitiated)`, ensuring consistent queue usage and easier testing. (`Sources/FeatureFlags.swift`, [[1]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4L352-R382) [[2]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4L361-R391) [[3]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4L428-R476) [[4]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4L518-R584) [[5]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4L882-R948)

**Synchronous Method Safety and Documentation:**

* Updated documentation for all synchronous flag retrieval methods (e.g., `getVariantSync`, `getAllVariantsSync`, etc.) to clearly state that these methods may block the calling thread and should not be called from the main UI thread. Also clarified fallback behavior when flags are not ready. (`Sources/FeatureFlags.swift`, [[1]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R178-R181) [[2]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R210-R213) [[3]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R241-R244) [[4]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R272-R275)
* Added runtime warnings (in debug builds) when synchronous flag retrieval methods are called from the main thread and a tracking queue is set, to help developers avoid UI-blocking calls. (`Sources/FeatureFlags.swift`, [[1]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R409-R426) [[2]](diffhunk://#diff-e13b1afbb6108893e0e3a7c5d9816b471e7352d4ea12e92740a8dd9c560b6fb4R558-R575)

### DIFF
--- a/Sources/FeatureFlags.swift
+++ b/Sources/FeatureFlags.swift
@@ -179,6 +179,8 @@ public protocol MixpanelFlags {
   ///   It is NOT recommended to call this from the main UI thread.
   ///   If flags are not ready (`areFlagsReady()` is false), this method returns the `fallback`
   ///   value, but it may still block while waiting for queued tracking or activation work to complete.
+  ///   If called immediately after track(), variants may not be activated yet due to a
+  ///   race condition as track is executed asynchronously. Use `getVariant` instead.
   ///
   /// - Parameters:
   ///   - flagName: The unique identifier for the feature flag.
@@ -212,6 +214,8 @@ public protocol MixpanelFlags {
   ///   It is NOT recommended to call this from the main UI thread.
   ///   If flags are not ready (`areFlagsReady()` is false), this method returns the `fallbackValue`,
   ///   but it may still block while waiting for queued tracking or activation work to complete.
+  ///   If called immediately after track(), variants may not be activated yet due to a
+  ///   race condition as track is executed asynchronously. Use `getVariantValue` instead.
   ///
   /// - Parameters:
   ///   - flagName: The unique identifier for the feature flag.
@@ -277,6 +281,8 @@ public protocol MixpanelFlags {
   ///   If flags are not ready (`areFlagsReady()` is false), it returns an empty dictionary
   ///   immediately without fetching, but it may still block while waiting for queued tracking
   ///   or activation work to complete.
+  ///   If called immediately after track(), variants may not be activated yet due to a
+  ///   race condition as track is executed asynchronously. Use `getAllVariants` instead.
   ///
   /// - Returns: A dictionary mapping flag names to their `MixpanelFlagVariant` values,
   ///            or an empty dictionary if flags are not ready.
@@ -355,11 +361,11 @@ class FeatureFlagManager: MixpanelFlags {
   private var trackingQueue: DispatchQueue
 
   // Initializers
-    internal init(serverURL: String, delegate: MixpanelFlagDelegate? = nil, trackingQueue: DispatchQueue) {
+    internal init(serverURL: String, trackingQueue: DispatchQueue, delegate: MixpanelFlagDelegate? = nil) {
         self.serverURL = serverURL
+        self.trackingQueue = trackingQueue
         self.delegate = delegate
         self.flagContext = delegate?.getOptions().featureFlagOptions.context ?? [:]
-        self.trackingQueue = trackingQueue
     }
 
   // --- Public Methods ---
@@ -397,17 +403,12 @@ class FeatureFlagManager: MixpanelFlags {
   // --- Sync Flag Retrieval ---
 
   func getVariantSync(_ flagName: String, fallback: MixpanelFlagVariant) -> MixpanelFlagVariant {
-    #if DEBUG
     if Thread.isMainThread {
       MixpanelLogger.warn(
         message: "It is NOT recommended to call this method from the main thread as it might block the calling thread until the value can be retrieved. Consider using async getVariant() instead."
       )
     }
-    #endif
-
-      return trackingQueue.sync {
-        return _getVariantSyncImpl(flagName, fallback: fallback)
-      }
+      return _getVariantSyncImpl(flagName, fallback: fallback)
   }
 
   private func _getVariantSyncImpl(_ flagName: String, fallback: MixpanelFlagVariant) -> MixpanelFlagVariant {
@@ -494,9 +495,7 @@ class FeatureFlagManager: MixpanelFlags {
           // This completion runs *after* fetch completes (or fails)
           let result: MixpanelFlagVariant
           if success {
-            // Fetch succeeded – call the private impl directly to avoid re-entering
-            // trackingQueue.sync (which would block the main thread unnecessarily
-            // and fire a false positive DEBUG warning).
+            // Fetch succeeded – call the private impl directly to avoid false positive DEBUG warning
             result = self._getVariantSyncImpl(flagName, fallback: fallback)
           } else {
             MixpanelLogger.warn(message: "Failed to fetch flags, returning fallback for \(flagName).")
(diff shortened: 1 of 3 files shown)


Sources/FeatureFlags.swift
```diff
@@ -341,6 +360,17 @@ class FeatureFlagManager: Network, MixpanelFlags {
     super.init(serverURL: serverURL)
   }
 
+  convenience init(serverURL: String, trackingQueue: DispatchQueue?) {
```
tylerjroach: Why would we allow trackingQueue to be nil? It is not optional on Mixpanel API

Sources/FeatureFlags.swift
```diff
@@ -349,7 +379,7 @@ class FeatureFlagManager: Network, MixpanelFlags {
 
   func loadFlags(completion: ((Bool) -> Void)?) {
     // Dispatch fetch trigger to allow caller to continue
-    DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+    getTrackingQueue().async { [weak self] in
```
tylerjroach: shouldn't need the getter if its required.

Sources/FeatureFlags.swift
```diff
@@ -955,12 +1021,14 @@ class FeatureFlagManager: Network, MixpanelFlags {
           // Track the feature flag check event with the new variant
           self._trackFlagIfNeeded(flagName: flagKey, variant: pendingEvent.pendingVariant)
 
-          // Record to backend (fire-and-forget)
-          self.recordFirstTimeEvent(
-            flagId: pendingEvent.flagId,
-            projectId: pendingEvent.projectId,
-            firstTimeEventHash: pendingEvent.firstTimeEventHash
-          )
+            DispatchQueue.global(qos: .userInitiated).async { [weak self] in
```
tylerjroach: MixpanelApi also has a network queue. Seems like it may be if we also passed it in for this purpose. We don't have to switch queue's here. We could ensure the queue is switched in the recordFirstTimeEventFunction. Although where the network call is actually made is already set up as non-blocking so I'm not even sure we need to switch here.

ketanmixpanel: We are not expecting any response back from this API call and by default its async. `DispatchQueue.global(qos: .userInitiated).async ` is used as it will do the API request creation on the background thread and the trackingQueue will return immediately.

Sources/FeatureFlags.swift
```diff
@@ -376,6 +406,24 @@ class FeatureFlagManager: Network, MixpanelFlags {
   // --- Sync Flag Retrieval ---
 
   func getVariantSync(_ flagName: String, fallback: MixpanelFlagVariant) -> MixpanelFlagVariant {
+    #if DEBUG
+    if Thread.isMainThread, trackingQueue != nil {
+      MixpanelLogger.warn(
+        message: "It is NOT recommended to call this method from the main thread as it might block the calling thread until the value can be retrieved. Consider using async getVariant() instead."
+      )
```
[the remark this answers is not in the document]
ketanmixpanel: @copilot apply changes based on [this feedback](https://github.com/mixpanel/mixpanel-swift/pull/715#discussion_r3094981231)

Sources/FeatureFlags.swift
```diff
@@ -341,6 +360,17 @@ class FeatureFlagManager: Network, MixpanelFlags {
...
+  convenience init(serverURL: String, trackingQueue: DispatchQueue?) {
+    self.init(serverURL: serverURL)
+    self.trackingQueue = trackingQueue
+  }
+
+  // --- Helper Methods ---
+
+  private func getTrackingQueue() -> DispatchQueue {
+    return trackingQueue ?? DispatchQueue.global(qos: .userInitiated)
+  }
```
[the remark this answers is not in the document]
ketanmixpanel: @copilot apply changes based on [this feedback](https://github.com/mixpanel/mixpanel-swift/pull/715#discussion_r3094981136)

Sources/FeatureFlags.swift
```diff
@@ -329,17 +351,16 @@ class FeatureFlagManager: Network, MixpanelFlags {
...
+  // Queue for synchronizing flag operations with tracking
+  private var trackingQueue: DispatchQueue
 
-  public init(serverURL: String, delegate: MixpanelFlagDelegate?) {
-    self.delegate = delegate
-    self.flagContext = delegate?.getOptions().featureFlagOptions.context ?? [:]
-    super.init(serverURL: serverURL)
-  }
+  // Initializers
+    internal init(serverURL: String, delegate: MixpanelFlagDelegate? = nil, trackingQueue: DispatchQueue) {
```
tylerjroach: since this is internal and we can change api, write optional parameters last

Sources/FeatureFlags.swift
```diff
@@ -376,6 +397,20 @@ class FeatureFlagManager: Network, MixpanelFlags {
   // --- Sync Flag Retrieval ---
 
   func getVariantSync(_ flagName: String, fallback: MixpanelFlagVariant) -> MixpanelFlagVariant {
+    #if DEBUG
```
tylerjroach: if DEBUG.doesn't work for customers, right since we ship release framework? We can still warn though if we detect main thread, but we don't need to try to gate it to debug builds.

Sources/FeatureFlags.swift
```diff
@@ -376,6 +397,20 @@ class FeatureFlagManager: Network, MixpanelFlags {
   // --- Sync Flag Retrieval ---
 
   func getVariantSync(_ flagName: String, fallback: MixpanelFlagVariant) -> MixpanelFlagVariant {
```
tylerjroach, line 405: I thought we determined to leave this as is. In its old form, it didn't do async. Here we can inform the customer of the race condition since track events are queued.

Sources/FeatureFlags.swift
ketanmixpanel: DEBUG will work here. We ship compiled framework for the Session Replay SDK.

Sources/FeatureFlags.swift
ketanmixpanel: Updating to run on caller thread.

tylerjroach (approved):

PR #721, opened 2026-04-30
fix(analytics): reload feature flags after MixpanelInstance.reset()

tylerjroach:
MixpanelInstance.reset() clears persisted identity and regenerates an anonymous distinct id, but never touches the FeatureFlagManager. As a result the cached `flags` dictionary survives the reset, and any subsequent getVariantValue / getVariant call returns stale pre-reset variants without going to the network. There is also no path from reset() to loadFlags() — reset() does not call identify(), which is where the existing loadFlags-on-id-change branch lives.

This change:

- Adds FeatureFlagManager.reset(), which posts to the trackingQueue and clears flags, trackedFeatures, pendingFirstTimeEvents, pendingFirstTimeEventNames, activatedFirstTimeEvents, and fetch timing under the flagsLock. Pending fetch-completion handlers are drained with `false`.
- Introduces a fetchGeneration counter snapshotted at the start of _performFetchRequest and re-checked in the success/failure closures (and once more under the lock before applying results) so an in-flight fetch dispatched before reset() does not overwrite the freshly cleared cache with results from the previous identity.
- Wires MixpanelInstance.reset() to call FeatureFlagManager.reset() and, if prefetchFlags is enabled, kick off a new loadFlags() under the new identity. The cast to FeatureFlagManager mirrors the existing pattern in Track.swift for internal flag-manager methods that don't belong on the public MixpanelFlags protocol.

### DIFF
--- a/Sources/FeatureFlags.swift
+++ b/Sources/FeatureFlags.swift
@@ -392,9 +392,17 @@ class FeatureFlagManager: MixpanelFlags {
   /// in-flight fetch dispatched before this call is discarded when it completes (via the
   /// generation check in `_performFetchRequest`'s success/failure closures). Pending
   /// fetch-completion handlers are invoked with `false` so callers don't hang.
-  func reset() {
+  ///
+  /// `completion` (optional) fires on the tracking queue once the in-memory state has been
+  /// cleared, so callers (notably `MixpanelInstance.reset(completion:)`) can defer their own
+  /// completion until after the wipe — without it, a customer calling a sync flag API from
+  /// the outer reset completion handler could observe pre-reset variants.
+  func reset(completion: (() -> Void)? = nil) {
     trackingQueue.async { [weak self] in
-      guard let self = self else { return }
+      guard let self = self else {
+        completion?()
+        return
+      }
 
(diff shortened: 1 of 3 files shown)


Sources/FeatureFlags.swift
```diff
@@ -381,6 +385,39 @@ class FeatureFlagManager: MixpanelFlags {
...
   }
 
+  /// Clears all in-memory feature flag state: cached flags, tracked-flag set, fetch timing,
+  /// and first-time event state. Intended to be called from `MixpanelInstance.reset()`.
+  ///
+  /// Posts to the tracking queue so the mutation is serialized with reads and fetches. Any
+  /// in-flight fetch dispatched before this call is discarded when it completes (via the
+  /// generation check in `_performFetchRequest`'s success/failure closures). Pending
+  /// fetch-completion handlers are invoked with `false` so callers don't hang.
+  func reset() {
```
ketanmixpanel: Lets add tests to verify the state reset and setContext() survives reset().

ketanmixpanel (approved):
LG

PR #722, opened 2026-05-04
Reset flags instance

ketanmixpanel:
This pull request improves the handling of feature flag state resets in the `MixpanelInstance` class to ensure pending fetch completion handlers are properly managed and notified when a reset occurs. The changes add a mechanism to drain and clear pending handlers before replacing the feature flag manager, preventing race conditions and ensuring correct notification.

Feature Flag Reset Improvements:

* Added a `drainCompletionHandlers()` method to `FeatureFlagManager` to extract and clear all pending fetch completion handlers before the instance is deallocated, allowing for proper notification of these handlers during a reset.
* Introduced a new `resetFeatureFlags()` method in `MixpanelInstance` that:
  * Drains pending completion handlers from the old flag manager,
  * Replaces the feature flag manager with a fresh instance (resetting all internal state and eliminating race conditions),
  * Notifies any pending handlers that their fetch will not complete due to the reset,
  * Triggers flag prefetching if configured.
* Updated the main reset logic in `MixpanelInstance` to call `resetFeatureFlags()` as part of the reset process, ensuring feature flag state and handlers are cleared and managed correctly.

### DIFF
--- a/Sources/FeatureFlags.swift
+++ b/Sources/FeatureFlags.swift
@@ -739,6 +739,24 @@ class FeatureFlagManager: MixpanelFlags {
     }
   }
 
+  /// Extracts and clears all pending fetch completion handlers before instance deallocation.
+  /// Called from MixpanelInstance.resetFeatureFlags() to retrieve handlers that need to be
+  /// notified before the instance is replaced.
+  ///
+  /// - Returns: Array of completion handlers that were pending. The caller is responsible
+  ///   for invoking them with the appropriate result.
+  func drainCompletionHandlers() -> [(Bool) -> Void] {
+    var handlers: [(Bool) -> Void] = []
+
+    flagsLock.write {
+      handlers = self.fetchCompletionHandlers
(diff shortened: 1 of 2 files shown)


tylerjroach:
Prefer #721 instead of this one for the reset fix, even after the discard commit was applied to this PR.

The "replace the entire FeatureFlagManager" pattern in this PR is conceptually appealing.

The problem surfaced while planning the variant-cache feature on top: not every field on FeatureFlagManager should reset on identity change. flagContext (anything set via setContext()) is a runtime override that should survive identify() and reset(). The "throw away the instance" pattern silently drops it because the new instance re-derives flagContext from options.featureFlagOptions.context (the init-time value).

You can fix it by snapshotting flagContext from the old instance and seeding the new one via a new init parameter. That works, but it inverts the maintenance burden rather than removing it: instead of "remember to clear new state in reset()" (#721), it becomes "remember to carry forward new state in resetFeatureFlags()" (this PR).

PR #723, opened 2026-05-05
feat: Feature Flag persistence

tylerjroach:
<h2>Configuration types</h2>
<h3><code>VariantLookupPolicy</code></h3>
<p>Idiomatic Swift enum with associated values — TTL uses <code>TimeInterval</code> (seconds) to match Foundation conventions.</p>
<pre><code class="language-swift">public enum VariantLookupPolicy {
  case networkOnly
  case cacheFirst(ttl: TimeInterval)
  case networkFirst(ttl: TimeInterval)
}
</code></pre>
<h3><code>FeatureFlagOptions</code> — new property</h3>
<pre><code class="language-swift">public let variantLookupPolicy: VariantLookupPolicy   // default: .networkOnly
</code></pre>
<p>Caching behavior is derived directly from the policy — there is no separate <code>cacheVariants</code> toggle:</p>
<ul>
<li><code>.networkOnly</code> — no caching. The on-disk blob is also wiped on init if present, so toggling from a caching policy back to <code>.networkOnly</code> cleans up after itself.</li>
<li><code>.cacheFirst(ttl:)</code> / <code>.networkFirst(ttl:)</code> — successful fetches write to disk; the cache is read on init.</li>
</ul>
<h3><code>MixpanelFlagVariant.Source</code> (on served variants)</h3>
<p>Discriminated union surfaced via the new <code>MixpanelFlagVariant.source</code> field. <code>nil</code> on developer-supplied fallbacks; otherwise:</p>
<pre><code class="language-swift">extension MixpanelFlagVariant {
  public enum Source {
    case network                  // variant came from /flags/
    case cache(cachedAt: Date)          // variant came from on-disk cache
  }
}
</code></pre>
<p>The cache timestamp lives on the <code>.cache</code> case (not as a parallel field on <code>MixpanelFlagVariant</code>), so invalid combinations like <code>.network</code> + non-nil timestamp are unrepresentable.</p>
<h2>Behavior matrix</h2>

variantLookupPolicy | Cache writes | Cache reads on init | Async lookup | Sync lookup / areFlagsReady()
-- | -- | -- | -- | --
.networkOnly | no | no (and wipes any stale blob from a prior caching session) | network only (today's behavior) | as today
.cacheFirst(ttl:) | yes | yes | serve cached immediately, refresh in background | reflects cache
.networkFirst(ttl:) | yes | yes | await network; on failure serve cached | reflects cache

<h2>Customer usage examples</h2>
<h3>Default (no caching — current behavior, no opt-in needed)</h3>
<pre><code class="language-swift">let options = MixpanelOptions(
    token: TOKEN,
    featureFlagOptions: FeatureFlagOptions(enabled: true)
)
let mixpanel = Mixpanel.initialize(options: options)
</code></pre>
<h3>CacheFirst — consistency-across-sessions over freshness</h3>
<p>Serves cached variants immediately on app start (within TTL), refreshes from network in the background.</p>
<pre><code class="language-swift">let flags = FeatureFlagOptions(
    enabled: true,
    variantLookupPolicy: .cacheFirst(ttl: 24 * 60 * 60)
)

let options = MixpanelOptions(
    token: TOKEN,
    featureFlagOptions: flags
)
let mixpanel = Mixpanel.initialize(options: options)
</code></pre>
<h3>NetworkFirst — freshness with offline resilience</h3>
<p>Async lookups await the network call; only fall back to cached variants if it fails.</p>
<pre><code class="language-swift">let flags = FeatureFlagOptions(
    enabled: true,
    variantLookupPolicy: .networkFirst(ttl: 24 * 60 * 60)
)
</code></pre>
<h3>Inspecting where a served variant came from</h3>
<p><code>variant.source</code> is set by the SDK on every served (non-fallback) variant. Pattern matching makes the dispatch clean:</p>
<pre><code class="language-swift">mixpanel.flags.getVariant(&quot;checkout-experiment&quot;, fallback: fallback) { variant in
    switch variant.source {
    case .cache(let cachedAt):
        print(&quot;Served from cache, written at \\(cachedAt)&quot;)
    case .network:
        print(&quot;Served from a fresh /flags/ response&quot;)
    case nil:
        print(&quot;Served the developer-supplied fallback&quot;)
    }
}
</code></pre>
<h3>Sync lookups and <code>areFlagsReady()</code></h3>
<p>Both reflect cached values when a caching policy is configured. <code>areFlagsReady()</code> returns <code>true</code> once flags are in memory, regardless of whether they were loaded from network or cache — check <code>variant.source</code> if you need to distinguish.</p>

### DIFF
--- a/Sources/FeatureFlags.swift
+++ b/Sources/FeatureFlags.swift
@@ -40,6 +40,21 @@ public struct MixpanelFlagVariant: Decodable {
   public let isExperimentActive: Bool? // Corresponds to 'is_experiment_active' from API
   public let isQATester: Bool? // Corresponds to 'is_qa_tester' from API
 
+  /// Where this variant was sourced from. `nil` on developer-supplied fallback instances;
+  /// `.network` or `.cache(at:)` when the SDK serves a variant. For cached variants, the
+  /// timestamp lives on the `.cache` case so invalid combinations like "network with a
+  /// timestamp" are unrepresentable.
+  public let source: Source?
+
+  /// Identifies where a served variant came from.
+  public enum Source {
+    /// Variant assigned by the most recent successful `/flags/` network call.
+    case network
+    /// Variant loaded from the on-disk cache. `at` is the time the variant set was
+    /// originally written to the cache.
+    case cache(at: Date)
+  }
+
   enum CodingKeys: String, CodingKey {
     case key = "variant_key"
     case value = "variant_value"
@@ -63,6 +78,7 @@ public struct MixpanelFlagVariant: Decodable {
     experimentID = try container.decodeIfPresent(String.self, forKey: .experimentID)
     isExperimentActive = try container.decodeIfPresent(Bool.self, forKey: .isExperimentActive)
     isQATester = try container.decodeIfPresent(Bool.self, forKey: .isQATester)
+    source = nil
   }
 
   // Helper initializer with fallbacks, value defaults to key if nil
@@ -76,6 +92,36 @@ public struct MixpanelFlagVariant: Decodable {
     self.experimentID = experimentID
     self.isExperimentActive = isExperimentActive
     self.isQATester = isQATester
+    self.source = nil
+  }
+
+  /// Internal initializer used when stamping a served variant with its origin.
+  internal init(
+    key: String,
+    value: Any?,
+    experimentID: String?,
+    isExperimentActive: Bool?,
+    isQATester: Bool?,
+    source: Source?
+  ) {
+    self.key = key
+    self.value = value
+    self.experimentID = experimentID
+    self.isExperimentActive = isExperimentActive
+    self.isQATester = isQATester
+    self.source = source
+  }
+
+  /// Returns a copy of this variant stamped with the given source. Other fields are preserved.
+  internal func withSource(_ source: Source) -> MixpanelFlagVariant {
+    return MixpanelFlagVariant(
+      key: self.key,
+      value: self.value,
+      experimentID: self.experimentID,
+      isExperimentActive: self.isExperimentActive,
+      isQATester: self.isQATester,
+      source: source
+    )
   }
 }
 
@@ -162,10 +208,16 @@ public protocol MixpanelFlags {
   /// The completion handler is called with `true` on success and `false` on failure.
   func loadFlags(completion: ((Bool) -> Void)?)
 
-  /// Synchronously checks if the flags  have been successfully loaded
-  /// and are available for querying.
+  /// Synchronously checks if flag variants are in memory and available for synchronous access.
   ///
-  /// - Returns: `true` if the flags are loaded and ready for use, `false` otherwise.
+  /// - Note: When a caching `variantLookupPolicy` is configured (`.cacheFirst` or
+  ///   `.networkFirst`), this can return `true` before the SDK has spoken to the network this
+  ///   session — the returned variants may be stale data from a previous session. Use the
+  ///   `source` field on the served `MixpanelFlagVariant` to distinguish: `.network` for
+  ///   fresh values, `.cache(at:)` for on-disk-cache values (with the cache timestamp).
+  ///
+  /// - Returns: `true` if flag variants are available in memory (from network or cache),
+  ///            `false` otherwise.
   func areFlagsReady() -> Bool
 
   // --- Sync Flag Retrieval ---
@@ -300,9 +352,14 @@ public protocol MixpanelFlags {
 
   /// Replaces the current custom flag evaluation context entirely and triggers a flag re-fetch.
   ///
+  /// In-memory variants and the on-disk cache are intentionally NOT cleared — the cache is
+  /// keyed on distinctId only, so it remains valid for this user across context changes. The
+  /// next successful fetch under the new context will overwrite the cache with fresh values.
+  ///
   /// - Parameters:
   ///   - context: The new context dictionary to use for flag evaluation.
-  ///   - completion: A closure called when the fetch completes (success or failure).
+  ///   - completion: A closure called when the fetch under the new context completes (success
+  ///     or failure). Always invoked exactly once.
   func setContext(_ context: [String: Any], completion: @escaping () -> Void)
 }
 
@@ -331,6 +388,18 @@ class FeatureFlagManager: MixpanelFlags {
   var isFetching: Bool = false
   private var trackedFeatures: Set<String> = Set()
   private var fetchCompletionHandlers: [(Bool) -> Void] = []
+
+  /// True when `flags` was populated from the on-disk cache and we have not yet seen the
+  /// initial network response for the current user/context. Only set for `.networkFirst` —
+  /// `.cacheFirst` serves cached values immediately. Async lookups gate on this to honor the
+  /// NetworkFirst spec ("await on network call, only serve persisted values if it fails")
+  /// while still letting sync lookups + areFlagsReady() see the cached values.
+  internal var awaitingInitialNetworkResponse: Bool = false
+
+  /// Per-instance name used as the UserDefaults key prefix for the on-disk cache. Captured
+  /// from the delegate at init so we can persist/clear without holding a delegate reference
+  /// during async work.
+  private let instanceName: String
 
   // First-time event targeting state
   internal var pendingFirstTimeEvents: [String: PendingFirstTimeEvent] = [:]  // Keyed by "flagKey:firstTimeEventHash"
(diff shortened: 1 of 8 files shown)


Sources/FeatureFlags.swift
```diff
@@ -63,6 +78,7 @@ public struct MixpanelFlagVariant: Decodable {
     experimentID = try container.decodeIfPresent(String.self, forKey: .experimentID)
     isExperimentActive = try container.decodeIfPresent(Bool.self, forKey: .isExperimentActive)
     isQATester = try container.decodeIfPresent(Bool.self, forKey: .isQATester)
+    source = nil
```
ketanmixpanel: `source == nil` means value returned using fallback, then having one more case in Source named `fallback` is meaningful than returning nil for fallback. What do you think?

Sources/MixpanelPersistence.swift
ketanmixpanel: Can you check and remove the existing functionality? It seems to be marked as not currently used.

Sources/FeatureFlags.swift
```diff
@@ -772,13 +919,167 @@ class FeatureFlagManager: MixpanelFlags {
...
+        // Persist the raw response so future sessions / failed fetches can fall back. Writes
+        // are independent of the lookup policy — a customer may opt into caching just to warm
+        // up for a future migration to a caching policy. Skipped when `didApplyResults` is
+        // false because that means the generation check failed (reset raced ahead) and we
+        // shouldn't overwrite the cache with prior-user data. UserDefaults I/O is kept
+        // outside the lock since it can block.
+        if didApplyResults,
+           shouldCache,
+           let rawData = rawHolder.data,
+           let rawString = String(data: rawData, encoding: .utf8) {
```
ketanmixpanel, line 1015: Instead of converting the rawData to string, can you pass the rawData as is to FlagsCacheBlob to store it? This way we can save the string conversion for read and write both.

Sources/MixpanelOptions.swift
```diff
@@ -45,14 +47,60 @@ public struct FeatureFlagOptions {
...
+///   network in the background. Cached entries older than `ttl` are ignored on read but NOT
+///   deleted (the next successful fetch overwrites them; a longer TTL on a future launch
+///   could reuse them). Pass a non-positive TTL to effectively disable expiry.
+/// - `networkFirst(ttl:)`: Prefer fresh values from the network, but fall back to cached
+///   variants when the network call fails. Same TTL semantics as `cacheFirst`.
+///
+/// Convenience zero-argument forms `cacheFirst()` / `networkFirst()` use `defaultTTL`
+/// (1 hour) — equivalent to passing `ttl: VariantLookupPolicy.defaultTTL` explicitly.
+public enum VariantLookupPolicy {
+  case networkOnly
```
ketanmixpanel, line 100: Do you think networkOnly needs a ttl too? There is chance of hitting the stale network only variants if the app was put in background for longer time. 
If we consider the default TTL to be 1 hour, this is very much possible with network values.

Sources/FeatureFlags.swift
```diff
@@ -361,15 +430,40 @@ class FeatureFlagManager: MixpanelFlags {
...
+
+        // Dispatch the init-time cache work on the tracking queue so UserDefaults I/O doesn't
+        // block the caller (typically the main thread during MixpanelInstance construction).
+        // Caching policies load the cache; `.networkOnly` wipes any stale blob left over from
+        // a previous session that used a caching policy.
+        if let options = delegate?.getOptions(), options.featureFlagOptions.enabled {
+            switch options.featureFlagOptions.variantLookupPolicy {
+            case .cacheFirst, .networkFirst:
+                trackingQueue.async { [weak self] in
+                    self?._loadCachedVariants()
```
ketanmixpanel: UserDefaults I/O on tracking queue could block event tracking and other flags operations if cache is large. Can we move the disk operations to different thread?

Sources/FeatureFlags.swift
```diff
@@ -772,13 +919,167 @@ class FeatureFlagManager: MixpanelFlags {
...
+        if didApplyResults,
+           shouldCache,
+           let rawData = rawHolder.data,
+           let rawString = String(data: rawData, encoding: .utf8) {
+          let blob = FlagsCacheBlob(
+            cachedAt: fetchEndTime,
+            distinctId: distinctIdAtDispatch,
+            response: rawString
+          )
+          MixpanelPersistence.saveFlagsCache(blob, instanceName: self.instanceName)
```
ketanmixpanel: saveFlagsCache is called on the network thread (background) while loadFlagsCache is called on trackingQueue

Sources/MixpanelOptions.swift
tylerjroach: Interseting point, but even in the networkFirst scenario, the ttl is only for the cahce fallback. Network types don't have ttl, and we don't have a concept of an expiring network response. Fair point to bring up to the team.

Sources/FeatureFlags.swift
tylerjroach: UserDefaults is loaded into memory on app launch. This isn't a disk operation. The timing of this should be negligible. Realistic size of the payload is not going to take any meaningful time to load. Tracking queue may also make sense as the result is needed for track event lookups.

Other present queue here is network which doesn't seem to fit. 

Claude evaluation below:
> Saves: already not on tracking queue. Concern doesn't apply.
Init load: on tracking queue, but it's the very first work item — nothing's queued behind it yet.
Reset delete: on tracking queue intentionally, for atomicity with the in-memory wipe.
Could we use networkQueue? Technically yes, but it's named/scoped for event flushing. Adding flag I/O to it muddies the boundary. We'd more likely create a dedicated flagsIOQueue if we wanted isolation.

> My take: the perceived "blocking the tracking queue" concern doesn't really exist in practice — saves are already off it, init load is once-per-session and runs before anything stacks up, deletes are intentional.

Sources/FeatureFlags.swift
tylerjroach: Claude:

> My take: Intentional asymmetry, not a bug.
Each call site uses its current queue → no extra dispatch hop
UserDefaults is documented thread-safe
Making them consistent means either: routing saves through the tracking queue (adds latency, blocks tracking), or routing both through a new I/O queue (adds complexity)
The asymmetry is just "do the I/O where you are, since it's safe." Worth replying with "intentional — UserDefaults is thread-safe and we avoid the extra hop."

Sources/FeatureFlags.swift
tylerjroach: Claude:

> Current flow: Data → String → [String: Any] → JSONSerialization → Data → UserDefaults. On load, reverse plus another String → Data for JSONDecoder.

> Suggested: Skip the UTF-8 round-trip; store Data directly.

> My take: Real but tiny cost, and the fix has friction.

> The two UTF-8 conversions on a typical few-KB payload are microseconds
JSONSerialization can't hold raw Data in a dict, so a "real" fix means switching to PropertyListSerialization (which can) or splitting the response into its own UserDefaults key (loses atomicity)
Either fix is more invasive than the perf win justifies
I'd push back unless we have evidence this is measurably slow. Worth replying with "happy to do it but it's a few μs per fetch — want to keep the current shape unless we see real cost?"

Sources/MixpanelPersistence.swift
tylerjroach: Pushed

Sources/FeatureFlags.swift
```diff
@@ -603,23 +715,21 @@ class FeatureFlagManager: MixpanelFlags {
...
 
-      var currentFlags: [String: MixpanelFlagVariant]?
-      self.flagsLock.read { currentFlags = self.flags }
-      if let currentFlags = currentFlags {
-        DispatchQueue.main.async { completion(currentFlags) }
+      var canServeImmediately = false
+      self.flagsLock.read {
+        canServeImmediately = (self.flags != nil) && !self.awaitingInitialNetworkResponse
+      }
+      if canServeImmediately {
```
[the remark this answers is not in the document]
tylerjroach: This was a good find.

Sources/FeatureFlags.swift
```diff
@@ -503,20 +609,24 @@ class FeatureFlagManager: MixpanelFlags {
...
+      // caller falls through to the fetch path below.
       self.flagsLock.read {
-        flagsAreCurrentlyReady = (self.flags != nil)
-        if flagsAreCurrentlyReady, let currentFlags = self.flags {
-          if let variant = currentFlags[flagName] {
+        canServeImmediately = (self.flags != nil) && !self.awaitingInitialNetworkResponse
+        if canServeImmediately, let currentFlags = self.flags {
+          if let variant = currentFlags[flagName], !self.isVariantExpired(variant) {
             flagVariant = variant
             needsTrackingCheck = !self.trackedFeatures.contains(flagName)
```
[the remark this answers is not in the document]
tylerjroach: This was a good find.

ketanmixpanel (approved):
Verified the feature using the iOS sample app. Feature behavior confirmed working as expected.

PR #727, opened 2026-05-08
fix: global accent color issue

ketanmixpanel:
This pull request refactors how screen size is captured as an automatic property in the SDK. Instead of capturing screen size directly in the automatic properties initializer (which could cause threading issues), the logic is moved to a new `setUIProperties()` method that ensures screen size is captured on the main thread during SDK initialization. This improves thread safety and reliability, especially for iOS, tvOS, and macOS targets.

**Refactoring and Thread Safety Improvements:**

* Added `setUIProperties()` call to the `MixpanelInstance` initializer so that screen size is included in the `$ae_first_open` event and other automatic events.This pull request refactors how screen size is captured for automatic properties, primarily to avoid issues with SwiftUI initialization and potential deadlocks. Instead of capturing screen size synchronously during property initialization, the process is now deferred and performed asynchronously on the main thread. This ensures compatibility with SwiftUI and thread safety. The changes also introduce safeguards to ensure screen size is only captured once.

Linear: Fixes SDK-46


santigracia (dismissed):
LGTM!

Sources/AutomaticProperties.swift
```diff
@@ -94,6 +89,48 @@ class AutomaticProperties {
...
+  class func setUIProperties() {
+    #if os(iOS) || os(tvOS) || os(macOS)
+      DispatchQueue.main.async {
+          captureScreenSizeOnMainThread()
+      }
+    #endif
+  }
+
+  /// Internal method to capture screen size. Must be called on main thread.
+  private static func captureScreenSizeOnMainThread() {
```
tylerjroach: I know this method needs called on the main thread, but thats not what it does, so I don't think `OnMainThread` is appropriate. Is something like `@MainActor` appropriate? Or just move the DispatchQueue to this method.

ketanmixpanel: fixed.

tylerjroach (approved):

PR #737, opened 2026-06-09
feat: add backupHost support for failover (parity with Android SDK)

kaan-barmore-genc-mixpanel:
## Summary

- Adds `MixpanelOptions.backupHost` (default `nil`) and a settable `MixpanelInstance.backupHost` property, mirroring the Android SDK's `setBackupHost`
- On a flush request, if the primary `serverURL` fails with a connection error or non-4xx server error, retries once against the backup host (scheme/port preserved, only host swapped)
- 4xx client errors and parse errors do not trigger failover
- Scoped to the flush/ingestion path only (track/engage/groups); feature-flag fetches are intentionally untouched

## Design notes

- Failover is not applied to feature flags — deliberate, since this is about ingestion reliability
- `backupHost` accepts a bare host or a full URL (only the host is extracted). Android documents host-only; this is more lenient
- Retry count: one retry per flush attempt via the backup host. The existing flush loop already handles batch-level retries with backoff. Android does up to 3 host-pair attempts with progressive sleeps inside one request — if strict behavioral parity matters, this is the gap to revisit

## Test plan

- [x] 14 new tests in `MixpanelBackupHostTests` covering:
  - Host substitution helper (`BasePath.backupBaseURL`)
  - Fallback decision logic (`FlushRequest.shouldFallBackToBackup`)
  - Options -> instance -> flush -> request propagation
- [x] All 259 tests pass (1 pre-existing flaky failure in `MixpanelAutomaticEventsTests.testUpdated`, unrelated)

Resolves SDK-74


Sources/MixpanelOptions.swift
```diff
@@ -158,6 +158,21 @@ public class MixpanelOptions {
...
+  ///
+  /// This is intended for customers in regions where Mixpanel's primary endpoints are blocked:
+  /// configure a proxy and set its host here so that only traffic that actually fails over hits
+  /// the proxy, keeping the integration largely transparent. Mirrors the Android SDK's
+  /// `backupHost`.
+  ///
+  /// Provide a host name (e.g. `"api-backup.mixpanel.com"`). The primary URL's scheme and port
+  /// are preserved when failing over; only the host is swapped. A full URL is also accepted —
+  /// only its host is used.
+  public let backupHost: String?
```
ketanmixpanel: Let's not introduce one more config option to the `MixpanelOptions`. It will remain unused for most of the apps. Let's do it the way Android does it.  Events are persisted to SQLite DB immediately when tracked, and the retry mechanism with exponential backoff ensures they'll be sent once backupHost is set. So no data loss risk.

```
  let mixpanel = Mixpanel.initialize(token: "YOUR_TOKEN")                                                                   
  mixpanel.backupHost = "api-backup.mixpanel.com"    
```

ketanmixpanel (changes requested):

Sources/MixpanelOptions.swift
kaan-barmore-genc-mixpanel: Done.

MixpanelDemo/MixpanelDemoTests/MixpanelBackupHostTests.swift
```diff
@@ -0,0 +1,109 @@
...
+//  MixpanelDemoTests
+//
+//  Covers the backupHost failover feature (parity with the Android SDK's backupHost).
+//
+
+import XCTest
+
+@testable import Mixpanel
+
+class MixpanelBackupHostTests: MixpanelBaseTests {
```
ketanmixpanel, line 70: The current test suite in `MixpanelBackupHostTests.swift` has excellent coverage of the helper methods (URL substitution, fallback decision logic, property propagation), but is missing an integration test that verifies the backup host is actually called when the primary request fails.

Sources/FlushRequest.swift
```diff
@@ -83,22 +83,89 @@ class FlushRequest: Network {
...
+        // returned a non-4xx error — never for client errors (4xx), where a different host
+        // won't help.
+        if let backupBase = self.backupBaseURL(forPrimary: base, failureReason: reason) {
+          MixpanelLogger.info(
+            message:
+              "Primary host request to \(resource.path) failed (\(reason)); retrying via backup host"
+          )
+          Network.apiRequest(
+            base: backupBase, resource: resource,
+            failure: { (backupReason, _, backupResponse) in
```
ketanmixpanel: Add logs for success and failure with backup host for easy debugging.

Sources/FlushRequest.swift
```diff
@@ -83,22 +83,89 @@ class FlushRequest: Network {
...
+              self.handleFlushFailure(
+                path: resource.path, reason: backupReason, response: backupResponse,
+                completion: completion)
+            },
+            success: { (result, backupResponse) in
+              self.handleFlushSuccess(
+                base: backupBase, result: result, response: backupResponse, completion: completion)
+            })
+        } else {
+          self.handleFlushFailure(
```
ketanmixpanel: Add log for why backup host was not used.

Sources/FlushRequest.swift
kaan-barmore-genc-mixpanel: Added

Sources/FlushRequest.swift
kaan-barmore-genc-mixpanel: Done

ketanmixpanel (dismissed):

Sources/MixpanelOptions.swift
```diff
@@ -97,199 +97,201 @@ public struct FeatureFlagOptions {
...
+        // When featureFlagOptions is explicitly provided, it takes precedence
+        if let featureFlagOptions = featureFlagOptions {
+            self.featureFlagOptions = featureFlagOptions
+        } else {
+            self.featureFlagOptions = FeatureFlagOptions(
+                enabled: featureFlagsEnabled,
+                context: featureFlagsContext
+            )
+        }
     }
```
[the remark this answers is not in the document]
ketanmixpanel: the backup host will be set using the instance.backupHost instead of mixpanel options

Sources/FlushRequest.swift
```diff
@@ -9,121 +9,198 @@
...
+                        },
+                        success: { (result, backupResponse) in
+                            MixpanelLogger.info(
+                                message: "Backup host request to \(resource.path) succeeded")
+                            self.handleFlushSuccess(
+                                base: backupBase, result: result, response: backupResponse, completion: completion)
+                        })
+                } else {
+                    self.handleFlushFailure(
+                        path: resource.path, reason: reason, response: response, completion: completion)
```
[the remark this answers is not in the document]
ketanmixpanel: should be fine as same headers are replayed

rahul-mixpanel (approved):

PR #745, opened 2026-06-29
feat: extend Source.fallback with FallbackReason (SDK-79)

tylerjroach:
## Summary

- **`MixpanelFlagVariant.Source.fallback` now carries an associated `FallbackReason`**: `.flagNotFound` | `.notReady` | `.backendError`. `getVariantSync` stamps `.notReady` when flags haven't loaded and `.flagNotFound` when the key is absent from the loaded set. Async `getVariant` upgrades to `.backendError` when a network fetch fails and no cached/persisted flags are available.
- `FallbackReason` conforms to `Equatable, Hashable, Sendable` for testing, Swift 6 concurrency, and downstream collection use.
- `MixpanelFlagVariant.withSource(_:)` is now `public` so callers (and the openfeature provider) can stamp source metadata.
- New public init lets callers construct variants with an explicit `Source`.

## ⚠ Breaking change

`Source.fallback` gained an associated value. External code that pattern-matches or switches on the case will not compile as-is:

```swift
// Before
switch variant.source { case .fallback: ... }
if case .fallback = variant.source { ... }
MixpanelFlagVariant(..., source: .fallback())

// After
switch variant.source { case .fallback(let reason): ... }   // or .fallback(_)
if case .fallback(_) = variant.source { ... }               // or bind the reason
MixpanelFlagVariant(..., source: .fallback(reason: .flagNotFound))
```

The default-associated-value trick (`case fallback(reason: FallbackReason = .unspecified)`) was tried and dropped in this branch — it lets `.fallback()` compile at construction sites but doesn't rescue exhaustive `switch` statements or `if case` patterns, so it was misleading half-coverage. Ripping the default off forces every callsite to name an explicit reason, which is the actionable behavior we want.

## Why

Today the OpenFeature wrapper (in [mixpanel-swift-openfeature](https://github.com/mixpanel/mixpanel-swift-openfeature)) collapses every fallback to `FLAG_NOT_FOUND`, sending callers debugging a default value chasing the flag name when the real cause was usually `.notReady` or a backend failure. The new `FallbackReason` lets the wrapper map each case to the spec-correct error code.

Discriminated union (`Source.fallback(reason:)`) rather than two flat fields because Swift's enums model this naturally; invalid states like "successful evaluation with a fallback reason" are unrepresentable — mirroring the existing `.persistence(persistedAt:)` case's design. Reason names align with the Android SDK-79 shape ([mixpanel-android#981](https://github.com/mixpanel/mixpanel-android/pull/981) uses `Source.Fallback` with a `Reason` field); the dynamic-typed SDKs (JS, PHP) use a sibling flat `fallback_reason` string field with matching values.

Cross-SDK fix tracked at [Linear SDK-79](https://linear.app/mixpanel/issue/SDK-79).

## Server-side note

`.flagNotFound` today still collapses "flag does not exist" with "user not in any rollout" because the `/flags` response can't tell them apart. This change preps the SDK to surface the distinction once the server returns it; no behavioral change for that case until then.

## Follow-up PR

[mixpanel-swift-openfeature](https://github.com/mixpanel/mixpanel-swift-openfeature) needs a matching update to dispatch on `Source.fallback`'s reason — that PR depends on this one shipping in a new `mixpanel-swift` release first (the openfeature package consumes the released SPM dep). It'll come right after release.

## Test plan

- [x] `swift build` succeeds
- [x] Three new `FeatureFlagPersistenceTests` cover the reason distinction end-to-end and pass locally on `iPhone 17 Pro`:
  - `testGetVariantSyncReturnsNotReadyWhenFlagsNeverLoaded`
  - `testGetVariantSyncReturnsFlagNotFoundWhenKeyMissing`
  - `testGetVariantAsyncReturnsBackendErrorWhenFetchFailsWithoutCache`
- [x] All 8 `FeatureFlagPersistenceTests` that flaked on the prior CI run pass locally in 0.125s total — no reproducible regression from this branch
- [ ] Wrapper PR pending base SDK release (Phase 2)

### DIFF
--- a/Sources/FeatureFlags.swift
+++ b/Sources/FeatureFlags.swift
@@ -70,11 +70,15 @@ public struct MixpanelFlagVariant: Decodable {
     /// cooperation — both surface as `.flagNotFound` for now. Future server changes
     /// can add a more specific reason without breaking callers that already switch
     /// on this enum.
-    public enum FallbackReason {
+    public enum FallbackReason: Equatable, Hashable, Sendable {
         /// Flag key was not present in the cache or network response.
         case flagNotFound
         /// Flags were not ready when the sync lookup happened.
         case notReady
+        /// Network fetch failed and no cached/persisted variant was available.
+        /// Only surfaced on the async `getVariant` path — `getVariantSync` cannot
+        /// distinguish "network error" from "flags never loaded."
+        case backendError
     }
 
     enum CodingKeys: String, CodingKey {
@@ -740,11 +744,20 @@ class FeatureFlagManager: MixpanelFlags {
                 // the fallback.
                 MixpanelLogger.debug(
                     message: "Flags not yet servable, attempting fetch for getVariant '\(flagName)'...")
-                self._fetchFlagsIfNeeded { _ in
+                self._fetchFlagsIfNeeded { fetchOk in
                     // Hop back to the tracking queue so we can read flags + run the tracking check
                     // atomically. Whether the fetch succeeded or not, _getVariantSyncImpl returns the
                     // variant from `flags` if present (persisted or network) else the fallback.
                     self.trackingQueue.async {
+                        if !fetchOk, !self.areFlagsReady() {
+                            // Fetch failed and we have no cached/persisted flags to fall back on —
+                            // distinguish this from "flags never loaded" (.notReady) so callers
+                            // (e.g., the OpenFeature wrapper) can surface a backend/general error
+                            // rather than a not-ready state.
+                            let result = fallback.withSource(.fallback(reason: .backendError))
+                            DispatchQueue.main.async { completion(result) }
+                            return
+                        }
                         let result = self._getVariantSyncImpl(flagName, fallback: fallback)
                         DispatchQueue.main.async { completion(result) }
                     }
--- a/Sources/FeatureFlags.swift
+++ b/Sources/FeatureFlags.swift
@@ -60,11 +60,7 @@ public struct MixpanelFlagVariant: Decodable {
         /// ready. Previously a Fallback meant only "not a real variant," which collapsed
         /// several distinct outcomes into one (SDK-79). Callers — especially the OpenFeature
         /// wrapper — can dispatch on the reason to surface the correct user-facing error code.
-        ///
-        /// `.unspecified` is the default for developer-constructed fallbacks; the SDK
-        /// stamps a more specific reason before returning so callers should rarely
-        /// observe it on a returned variant.
-        case fallback(reason: FallbackReason = .unspecified)
+        case fallback(reason: FallbackReason)
     }
 
     /// Why the SDK returned the developer fallback.
@@ -75,9 +71,6 @@ public struct MixpanelFlagVariant: Decodable {
     /// can add a more specific reason without breaking callers that already switch
     /// on this enum.
     public enum FallbackReason {
-        /// Developer-constructed default. The SDK stamps a more specific reason
-        /// before returning, so callers should rarely observe this value.
-        case unspecified
         /// Flag key was not present in the cache or network response.
         case flagNotFound
         /// Flags were not ready when the sync lookup happened.
@@ -110,7 +103,7 @@ public struct MixpanelFlagVariant: Decodable {
         // Decoded variants are immediately re-stamped via `withSource` before being placed in
         // `flags`, so the customer never observes `.fallback` here. Defaulting to `.fallback`
         // keeps `source` non-optional without needing a sentinel "unstamped" case.
-        source = .fallback()
+        source = .fallback(reason: .flagNotFound)
     }
 
     // Helper initializer with fallbacks, value defaults to key if nil
@@ -127,7 +120,7 @@ public struct MixpanelFlagVariant: Decodable {
         self.experimentID = experimentID
         self.isExperimentActive = isExperimentActive
         self.isQATester = isQATester
-        self.source = .fallback()
+        self.source = .fallback(reason: .flagNotFound)
     }
 
     /// Initializer that stamps a served (or fallback) variant with its origin.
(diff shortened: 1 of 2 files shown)


Sources/FeatureFlags.swift
```diff
@@ -6,133 +6,181 @@ import jsonlogic
...
+    } else {
+      self.value = key
     }
+    self.experimentID = experimentID
+    self.isExperimentActive = isExperimentActive
+    self.isQATester = isQATester
+    self.source = .fallback()
+  }
+
+  /// Initializer used by callers (and tests) who want to construct a variant with
```
ketanmixpanel: Looks like duplicate initializer. Initializer from line 153 Initializer does same job.

Sources/FeatureFlags.swift
```diff
@@ -491,933 +539,922 @@ class FeatureFlagManager: MixpanelFlags {
...
+        // when the immediate-serve path is satisfied by the persisted blob (not yet
+        // overwritten by a successful fetch), kick off a background fetch with no
+        // completion handler so we don't await it. Idempotent — `_fetchFlagsIfNeeded`
+        // dedupes via `isFetching`. Self-stops once the fetch clears
+        // `loadedBlobPersistedAt`.
+        if servingFromPersistedBlob,
+           case .persistenceUntilNetworkSuccess = self.resolvedLookupPolicy {
+          self._fetchFlagsIfNeeded(completion: nil)
+        }
+        let result = flagVariant ?? fallback
```
ketanmixpanel: Missing fallback reason stamping

```suggestion
 let result = flagVariant ?? fallback.withSource(.fallback(reason: .flagNotFound))
```

ketanmixpanel (changes requested):

Sources/FeatureFlags.swift
```diff
@@ -6,133 +6,181 @@ import jsonlogic
...
+  public enum FallbackReason {
+    /// Developer-constructed default. The SDK stamps a more specific reason
+    /// before returning, so callers should rarely observe this value.
+    case unspecified
+    /// Flag key was not present in the cache or network response.
+    case flagNotFound
+    /// Flags were not ready when the sync lookup happened.
+    case notReady
+    /// An exception occurred during evaluation.
+    case evaluationError
```
ketanmixpanel: `evaluationError` case is not used anywhere, was there any specific intention to add it?

tylerjroach:
Thanks for the review @ketanmixpanel — pushed `388b9aa` addressing all three threads (also rebased onto `master` for the swift-format conflict).

1. **Duplicate initializer (line 133):** dropped the duplicate, promoted the consolidated init to `public`. The remaining one takes `source` as the trailing parameter (matching field declaration order) with all other args optional defaulting to `nil`. `withSource` updated to use it.
2. **Missing fallback reason stamping (line 745):** stamped `.fallback(reason: .flagNotFound)` on the async fast-path — same treatment the sync impl already did in `_getVariantSyncImpl`. The async path was the gap.
3. **Unused `.evaluationError` case (line 87):** removed. It was speculative — Swift's cache-lookup path can't naturally surface this (the `BACKEND_ERROR` server-SDK case applies to per-call remote eval, which Swift doesn't do). Trivial to add back the moment a real call site needs it.

Let me know if anything else.

ketanmixpanel (dismissed):

Sources/FeatureFlags.swift
```diff
@@ -54,9 +54,34 @@ public struct MixpanelFlagVariant: Decodable {
...
+    /// cooperation — both surface as `.flagNotFound` for now. Future server changes
+    /// can add a more specific reason without breaking callers that already switch
+    /// on this enum.
+    public enum FallbackReason {
+        /// Developer-constructed default. The SDK stamps a more specific reason
+        /// before returning, so callers should rarely observe this value.
+        case unspecified
+        /// Flag key was not present in the cache or network response.
+        case flagNotFound
+        /// Flags were not ready when the sync lookup happened.
```
ketanmixpanel, line 76: I see `BACKEND_ERROR` present in the JS SDK, but not in the Swift SDK. How API error will be handled here?

Sources/FeatureFlags.swift
```diff
@@ -54,9 +54,34 @@ public struct MixpanelFlagVariant: Decodable {
...
+        case notReady
```
ketanmixpanel, line 77: Does `unspecified` is equivalent to the `notReady` for our usecase here? I see that its just iOS and Android SDKs are using it, and other SDKs don't use it. 

As the value will be altered later without notice, can we use `notReady` as default? This way we will have the fallback reasons consistent across all other SDKs.

tylerjroach:
Thanks @ketanmixpanel — pushed `ceb0e72` addressing both of your comments and the two Greptile P2s. Summary of changes; details in the updated PR body.

**Ketan / BACKEND_ERROR (line 76):** added. `FallbackReason.backendError` is stamped on the async `getVariant` path when a fetch fails **and** no cached/persisted flags exist. `getVariantSync` stays as-is — it can't distinguish "network error" from "flags never loaded" (no network was attempted), so `.backendError` is async-only. New test `testGetVariantAsyncReturnsBackendErrorWhenFetchFailsWithoutCache` covers it.

**Ketan / unspecified vs notReady (line 77):** went further than the suggestion in `f3b13ca` — dropped `.unspecified` entirely and made `reason` required at every callsite. No placeholder can leak. Every value the customer observes is now actionable. Matches the Android SDK-79 shape ([mixpanel-android#981](https://github.com/mixpanel/mixpanel-android/pull/981)).

**Greptile P2 (protocol conformances):** accepted verbatim. `FallbackReason` now conforms to `Equatable, Hashable, Sendable`.

**Greptile P2 (no reason coverage):** added three tests in `FeatureFlagPersistenceTests` — `.notReady` on unloaded flags, `.flagNotFound` on loaded-but-missing key, `.backendError` on the async fetch-failure path.

**Greptile P1 (source-breaking change):** acknowledged, not reverting — the union matches Android and keeps invalid states unrepresentable (like the existing `.persistence(persistedAt:)` case). Called out prominently in the updated PR body with before/after examples so adopters aren't surprised. One factual footnote: Swift 5.1+ does support default associated values (`case fallback(reason: FallbackReason = .unspecified)` compiled fine) — but Greptile's core point stands: defaults help construction sites, not `switch` exhaustiveness. Since half-coverage was misleading, I dropped the default entirely.

**On the prior CI failure:** the 8 `FeatureFlagPersistenceTests` timeouts on `f3b13ca` were simulator flakes. All 8 pass locally in 0.125s total against `iPhone 17 Pro,OS=latest`. The `388b9aa` → `f3b13ca` diff was purely mechanical (enum case removal + explicit `.flagNotFound` at two placeholder sites) with no semantic mechanism for async timing regression. This push re-triggers CI; if it flakes on the same tests we can look into it separately.

ketanmixpanel (approved):

PR #746, opened 2026-06-29
feat: add autocapture frustration signals (click, rage click, dead click)

rahul-mixpanel:
## Summary

Adds automatic capture of user frustration signals on iOS:

- **Click tracking** (`$mp_click`): Captures tap events with semantic context (element ID, tag name, role, accessibility label, view hierarchy)
- **Rage click detection** (`$mp_rage_click`): Detects rapid repeated taps (4+ within 1s in 44pt radius)
- **Dead click detection** (`$mp_dead_click`): Identifies taps on interactive elements that produce no UI response within 500ms

Autocapture is **disabled by default** — opt-in via `AutocaptureOptions` in `MixpanelOptions`.

### Privacy

Autocapture does **not** capture visible text content (`$el_text`) from tapped elements. Tracking text can be invasive and raise privacy concerns. The complexity of nested view hierarchies can also cause text extraction to capture content from unintended views — for example, tapping a container view might extract text from a deeply nested label that isn't semantically related to the tap.

The captured properties (`$el_id`, `$el_tag_name`, `$attr-aria-label`, `$attr-role`, `$elements`, `$x`, `$y`) are purely structural UI metadata with no PII risk.

### Captured Properties

| Property | Description |
|----------|-------------|
| `$x`, `$y` | Touch coordinates |
| `$el_id` | Element identifier (accessibilityIdentifier → accessibilityLabel → fallback hash) |
| `$el_tag_name` | Element class name (e.g., `UIButton`, `Button`) |
| `$attr-aria-label` | Accessibility label |
| `$attr-role` | Semantic role (e.g., `Button`, `Switch`, `Slider`) |
| `$elements` | View hierarchy string (max 5 levels) |

### Architecture

| Component | Responsibility |
|-----------|---------------|
| `TouchInterceptor` | Touch interception via `UIApplication.sendEvent(_:)` method swizzling |
| `SemanticExtractor` | Element identification, role detection, hierarchy building |
| `RageClickTracker` | 4+ taps within 1000ms / 44pt radius |
| `DeadClickDetector` | No UI change within 500ms after tap on interactive element |
| `AutocaptureManager` | Lifecycle coordination, event dispatch |

### Dead Click Exclusions

Controls with inherent visual feedback are excluded from dead click monitoring:
UISwitch, UISlider, UITextField, UITextView, UIStepper, UISegmentedControl, UIDatePicker, UIPickerView (and their SwiftUI equivalents)

### Tests

- 7 UIKit autocapture instrumented tests (`AutocaptureUIKitInstrumentedTests`)
- SwiftUI autocapture instrumented tests (`AutocaptureSwiftUIInstrumentedTests`)
- Unit tests for ClickEvent model (`AutocaptureTests`)
- All autocapture tests pass on simulator (iOS 18.5)

### Demo app

- UIKit and SwiftUI test screens added to MixpanelDemo for manual validation

### Code Review Fixes

- Exclude controls with inherent feedback from dead click detection
- SwiftUI detection via class name patterns
- Depth-limited hierarchy walks

## Test plan

- [ ] Run all tests: `xcodebuild test -project MixpanelDemo/MixpanelDemo.xcodeproj -scheme MixpanelDemo -destination 'platform=iOS Simulator,name=iPhone 16 Pro' -only-testing:MixpanelDemoTests`
- [ ] Manual validation using UIKit and SwiftUI test screens in MixpanelDemo
- [ ] Verify existing SDK tests still pass


docs/autocapture.md
```diff
@@ -0,0 +1,310 @@
...
+```swift
+// UIKit
+sensitiveView.accessibilityIdentifier = "mp-sensitive"
+
+// SwiftUI
+SecretView()
+    .accessibilityIdentifier("mp-sensitive")
+```
+
+### Using accessibilityLabel
```
ketanmixpanel: We should add this option of marking view as sensitive to Session replay as well. This way you just mark a view as sensitive using `accessibilityLabel` and it will take effect everywhere.

docs/autocapture.md
```diff
@@ -0,0 +1,310 @@
...
+```swift
+// UIKit
+sensitiveView.accessibilityLabel = "mp-no-track"
+
+// SwiftUI
+SecretView()
+    .accessibilityLabel("mp-no-track")
+```
+
+Both `mp-sensitive` and `mp-no-track` are supported. The check uses `contains()`, so identifiers like `payment-form-mp-sensitive` also work.
```
ketanmixpanel: why do we have `mp-sensitive` and `mp-no-track` two identifiers ? cant we just have one?

rahul-mixpanel: The web sdk uses it for different purposes. I think for mobile. we just need one. Will keep mp-no-track.

Sources/MixpanelInstance.swift
ketanmixpanel, line 554: If user has opted out of tracking then the will `autocapture` remain disabled? If yes, then I think autocapture initialization should be done after checking the `optOutTrackingByDefault`.

MixpanelDemo/MixpanelDemo/AppDelegate.swift
```diff
@@ -136,10 +136,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         // ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
+        // Enable autocapture for testing
+        let autocaptureOptions = AutocaptureOptions()
+
         let mixpanelOptions = MixpanelOptions(
-            token: "MIXPANEL_TOKEN",
+            token: "3124becc8e700e259476cca2f895e5f5",
```
ketanmixpanel: remove token

Sources/MixpanelOptions.swift
```diff
@@ -253,6 +253,68 @@ public class MixpanelOptions {
...
+    /// )
+    /// ```
+    ///
+    /// **Note:** Autocapture is only available on iOS.
+    #if os(iOS)
+    public let autocaptureOptions: AutocaptureOptions?
+    #endif
+
+    #if os(iOS)
+    public init(
```
ketanmixpanel, line 275: With having multiple inits we will endup with same issue as what happened with original mixpanel initializer. The MixpanelOption is created with constants configs, and ideally it should have been variables allowing app developer to modify/update it after creating the options object.  One we add one more init, there is no way to revert it. Lets discuss it our sync on this.

Mixpanel-swift.podspec
```diff
@@ -42,7 +42,7 @@ Pod::Spec.new do |s|
   }
 
   s.subspec 'Complete' do |ss|
-    ss.ios.source_files = ['Sources/*.swift']
+    ss.ios.source_files = ['Sources/*.swift', 'Sources/Autocapture/**/*.swift']
```
ketanmixpanel: Lets include all swift files from the Sources folder -

```suggestion
    ss.ios.source_files = ['Sources/**/*.swift']
```

Sources/Autocapture.swift
```diff
@@ -53,25 +48,78 @@ open class Autocapture {
...
     open func trackScreenLeave(screenName: String, properties: Properties? = nil) {
-        guard let mixpanelInstance = mixpanelInstance else { return }
         guard !screenName.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
             MixpanelLogger.warn(
                 message: "trackScreenLeave called with empty screenName, ignoring event")
             return
         }
 
-        var mergedProperties: Properties = [:]
+        var mergedProperties: Properties = properties ?? [:]
```
ketanmixpanel: The redundant code from `trackScreenView` and `trackScreenLeave` can be moved to `trackAutocaptureEvent` with screenName param.

```
func trackAutocaptureEvent(_ eventName: String, screenName: String, properties: Properties) 
```

context/AUTOCAPTURE.md
```diff
@@ -0,0 +1,272 @@
+# iOS Autocapture
+
+Autocapture automatically tracks user interactions in your iOS app without requiring manual instrumentation.
```
ketanmixpanel, line 3: The `docs` folder will be overwritten during the release by generate docs script. The files that you have added will be deleted. If you want them to be present in the repo for reference, then you will have to move it to different location.

Sources/Autocapture/Model/ClickEvent.swift
```diff
@@ -0,0 +1,101 @@
...
+    /// - Parameters:
+    ///   - x: Touch location X coordinate
+    ///   - y: Touch location Y coordinate
+    ///   - elementId: Primary element identifier
+    ///   - tagName: Class name of the tapped view (defaults to empty string)
+    ///   - ariaLabel: Accessibility label (defaults to nil)
+    ///   - role: Semantic role of the element (defaults to nil)
+    ///   - elements: View hierarchy string (defaults to empty string)
+    public init(x: CGFloat, y: CGFloat, elementId: String,
+                tagName: String = "", ariaLabel: String? = nil,
```
ketanmixpanel: ariaLabel is very specific to the HTML, can we use native term `accessibilyLabel` instead?

Sources/Autocapture/Model/ClickEvent.swift
```diff
@@ -0,0 +1,101 @@
...
+                role: String? = nil, elements: String = "") {
```
ketanmixpanel: Any specific reason for defaulting tagName and elements to string and not to nil?

Sources/Autocapture/AutocaptureOptions.swift
```diff
@@ -0,0 +1,204 @@
...
+  #if os(iOS)
+    /// UIKit controls with inherent visual feedback that should be excluded from
+    /// dead click detection. These controls always produce a visible UI response
+    /// (toggle animation, keyboard, wheel picker, etc.) that may not be captured
+    /// by snapshot comparison.
+    static let excludedControlTypes: [AnyClass] = [
+      UISwitch.self,
+      UISlider.self,
+      UITextField.self,
+      UITextView.self,
```
ketanmixpanel: Check if these make sense to add to the excluded control types - Activity Indicator (loader), progress view, page control.

Sources/Autocapture/AutocaptureOptions.swift
```diff
@@ -0,0 +1,204 @@
...
+/// Dead clicks are detected when a user taps on an interactive element
+/// but no visible UI change occurs, indicating a broken or unresponsive element.
+public struct DeadClickOptions {
+  /// Whether dead click detection is enabled. Defaults to `true`.
+  public let enabled: Bool
+
+  /// Timeout in milliseconds to wait for UI response after a click.
+  /// If no UI change is detected within this window, a dead click is recorded.
+  /// Defaults to `500` (0.5 seconds).
+  public let timeoutMs: Int
```
ketanmixpanel: Shall we use `timeWindowMs` instead of `timeoutMs` similar to RageClickOptions?

Sources/Autocapture/AutocaptureOptions.swift
```diff
@@ -0,0 +1,204 @@
...
+        }
+      }
+      return false
+    }
+  #endif
+}
+
+// MARK: - Click Options
+
+/// Configuration options for basic click event capture.
```
ketanmixpanel: can you provide additional details on the click capture here? I feel basic click event capture is not sufficient.

Sources/Autocapture/DeadClickDetector.swift
```diff
@@ -0,0 +1,319 @@
...
+      currentWindow = window
+      pendingCheck = PendingCheck(
+        event: event,
+        baselineSnapshot: baseline,
+        startTime: Date()
+      )
+      lock.unlock()
+
+      // Schedule final check at full timeout
+      DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(timeoutMs)) {
```
ketanmixpanel: Can you convert this into a task and cancel the task if you receive a new click in the timeoutMS duration?
Currently its running the `performFinalCheck` on main thread all the time.

Sources/Autocapture/SemanticExtractor.swift
```diff
@@ -0,0 +1,230 @@
...
+    // MARK: - Interactive View Resolution
+
+    /// Check if a view is interactive (has tap handlers or is a UIControl with targets).
+    private func isInteractive(_ view: UIView) -> Bool {
+      return AutocaptureDefaults.isInteractive(view)
+    }
+
+    /// Walk up the view hierarchy to find the nearest interactive ancestor.
+    /// Returns nil if no interactive ancestor is found within maxDepth levels.
+    private func findInteractiveAncestor(of view: UIView, maxDepth: Int = 5) -> UIView? {
```
ketanmixpanel: Can you pass `AutocaptureDefaults.maxHierarchyDepth` for maxDepth  instead of using hard coded value?

Sources/Autocapture/SemanticExtractor.swift
```diff
@@ -0,0 +1,230 @@
...
+        // SwiftUI: accessibilityLabel is primary
+        if let label = findAccessibilityLabel(in: view), !label.isEmpty {
+          return label
+        }
+        // SwiftUI: accessibilityIdentifier only works with VoiceOver, check anyway
+        if let identifier = findAccessibilityIdentifier(in: view), !identifier.isEmpty {
+          return identifier
+        }
+      } else {
+        // UIKit: accessibilityIdentifier is primary
```
ketanmixpanel: If we are going to recommend using `AccessibilityLabel` then should it be first? Giving different recommendation for SwiftUI and UIKit can be confusing.

rahul-mixpanel: I was having similar thoughts. let us keep `AccessibilityLabel` first for both cases.

Sources/Autocapture/SemanticExtractor.swift
```diff
@@ -0,0 +1,230 @@
...
+    ) -> String {
+      var hierarchy: [String] = []
+      var currentView: UIView? = view
+      var level = 0
+
+      while let v = currentView, level < maxLevels {
+        var name = String(describing: type(of: v))
+
+        // Add identifier if available
+        if let identifier = v.accessibilityIdentifier, !identifier.isEmpty {
```
ketanmixpanel: are we not using `AccessibilityLabel` here?

Sources/Autocapture/TouchInterceptor.swift
```diff
@@ -0,0 +1,297 @@
...
+      // Don't change state - we're just observing
+    }
+
+    override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent) {
+      defer {
+        downLocation = nil
+        downTime = nil
+        // Transition to failed state to let the gesture system continue
+        // We're observing only, not claiming the touch
+        state = .failed
```
ketanmixpanel: The state should reset to `.ended`?

Sources/Autocapture/AutocaptureManager.swift
```diff
@@ -0,0 +1,177 @@
...
+
+      // Emit click event
+      if options.clickOptions.enabled {
+        autocapture?.trackClick(clickEvent)
+        MixpanelLogger.debug(
+          message: "AutocaptureManager: emitted $mp_click for \(clickEvent.elementId)")
+      }
+
+      // Emit rage click event (independent of regular click)
+      if options.rageClickOptions.enabled, rageClickResult?.isRageClick == true {
```
ketanmixpanel: `options.rageClickOptions.enabled` check is not needed here.

```suggestion
      if rageClickResult?.isRageClick == true {
```

Sources/Autocapture/AutocaptureOptions.swift
```diff
@@ -0,0 +1,204 @@
...
+  public let enabled: Bool
+
+  /// Timeout in milliseconds to wait for UI response after a click.
+  /// If no UI change is detected within this window, a dead click is recorded.
+  /// Defaults to `500` (0.5 seconds).
+  public let timeoutMs: Int
+
+  public init(
+    enabled: Bool = true,
+    timeoutMs: Int = 500
```
ketanmixpanel: 500ms UI snapshot comparison can produce false positives for legitimate UI patterns. I think 1 sec is a safe middle ground, especially for mobile where network-triggered UI updates can be slower. The app can always change it to 500 ms to tighten it.

Sources/Autocapture/AutocaptureManager.swift
```diff
@@ -0,0 +1,177 @@
...
+
+      // Extract semantic information
+      let clickEvent = semanticExtractor.extractSemantics(from: view, at: point)
+
+      // Check for rage click
+      let rageClickResult = rageClickTracker?.trackClick(x: point.x, y: point.y)
+
+      // Emit click event
+      if options.clickOptions.enabled {
+        autocapture?.trackClick(clickEvent)
```
ketanmixpanel: If a click is rage click, then shall we skip the regular click and just send rage click? for same click two events will be sent in current case

rahul-mixpanel: The JS SDK (mixpanel-js-private) sends both `$mp_click` and `$mp_rage_click` independently — they are separate listeners and rage click does not suppress the regular click. Our current mobile behavior matches JS. No change needed.

Sources/Autocapture/AutocaptureOptions.swift
rahul-mixpanel: Added `UIPageControl` — it's interactive (tappable) and has inherent visual feedback (dots change state).

Skipped `UIActivityIndicatorView` and `UIProgressView` — these are non-interactive display-only controls by default. They won't pass the `isInteractive` check (no action targets, no tap gesture recognizers, no .button accessibility trait), so they'll never enter dead click detection in the first place.

Sources/Autocapture/AutocaptureOptions.swift
```diff
@@ -0,0 +1,220 @@
...
+/// )
+///
+/// let options = MixpanelOptions(
+///     token: "YOUR_TOKEN",
+///     autocaptureOptions: autocaptureOpts
+/// )
+/// ```
+public struct AutocaptureOptions {
+  /// Configuration for basic click capture.
+  public let clickOptions: ClickOptions
```
ketanmixpanel: what do you think of making thse AutocaptureOptions as optional? If i want to enable only the clickOptions, I do not need to set the rageclickoptions and deadclickoptions to false. 
If I want to set them all, I will go ahead and create objects with config for it. During the init as well it should be optional with default to nil, so you can pass the objects that you want to enable and rest remains disabled.

context/AUTOCAPTURE.md
rahul-mixpanel: moved to /context folder

Sources/Autocapture/SemanticExtractor.swift
```diff
@@ -0,0 +1,214 @@
...
+      }
+      // accessibilityIdentifier as fallback
+      if let identifier = findAccessibilityIdentifier(in: view), !identifier.isEmpty {
+        return identifier
+      }
+
+      // Fallback: ClassName_<hex hash>
+      let className = String(describing: type(of: view))
+      let safeHash = view.hash == Int.min ? Int.max : abs(view.hash)
+      return "\(className)_\(String(safeHash, radix: 16))"
```
ketanmixpanel: just a thought, is it possible to make the safeHash consistent across the screen visits?

Sources/Autocapture/Model/ClickEvent.swift
```diff
@@ -0,0 +1,153 @@
...
+      if let tagName = tagName {
+        props["$el_tag_name"] = tagName
+      }
+
+      if let elements = elements {
+        props["$elements"] = elements
+      }
+
+      if let accessibleLabel = accessibleLabel {
+        props["$attr-aria-label"] = accessibleLabel
```
ketanmixpanel: can we use accessibility label instead of aria label here?

Sources/Autocapture/RageClickTracker.swift
```diff
@@ -0,0 +1,128 @@
...
+    let isRageClick: Bool
+  }
+
+  /// Tracks rapid repeated clicks to detect user frustration (rage clicks).
+  ///
+  /// A rage click is detected when a user taps multiple times within a short time window
+  /// in approximately the same location, indicating frustration with an unresponsive element.
+  ///
+  /// Default thresholds (matching JS SDK):
+  /// - 4 or more clicks
```
ketanmixpanel: If no of clicks at same location crosses the threshold, then we are sending the rage click. Once we send the rage click, then shall we reset the threshold? 

Whats happening now is if I am tapping at same location for 10 times, then 4 or more means it will send 6 rage click events.

Sources/Autocapture/Model/ClickEvent.swift
rahul-mixpanel: No. This is what JS SDK also sends. we're trying to align the param names with JS SDK here.

context/AUTOCAPTURE.md
```diff
@@ -0,0 +1,271 @@
...
+### ClickOptions
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `enabled` | `true` | Track all click events |
+
+### RageClickOptions
+
+| Option | Default | Description |
+|--------|---------|-------------|
```
ketanmixpanel, line 44: clicking back button multiple times may report a false rage click as the back buttons are located at same place and can be clicked multiple times based on the navigation stack.

ketanmixpanel: screen name can be used to avoid this rage click

Sources/Autocapture/TouchInterceptor.swift
```diff
@@ -0,0 +1,313 @@
...
+    deinit {
+      NotificationCenter.default.removeObserver(self)
+    }
+
+    // MARK: - Public API
+
+    /// Install the touch interceptor.
+    ///
+    /// - Parameter manager: The AutocaptureManager to delegate touch events to
+    func install(manager: AutocaptureManager) {
```
ketanmixpanel, line 42: Do you think a different name like setupTouchInterceptor and removeTouchInterceptor will be more meaningful than install and uninstall?

Sources/Autocapture/AutocaptureManager.swift
rahul-mixpanel: We discussed this and decided to keep sending both $mp_click and $mp_rage_click for the same tap. A rage click is still a click, and we want the click count to remain accurate. Filtering rage clicks out of regular clicks would undercount total interactions.

Sources/Autocapture/AutocaptureManager.swift
rahul-mixpanel: Addressed — the redundant `options.rageClickOptions.enabled` check has been removed. The code now directly checks `rageClickResult?.isRageClick == true`. The enabled flag is only used at init time to decide whether to create the tracker.

Sources/Autocapture/SemanticExtractor.swift
rahul-mixpanel: Addressed — `generateElementId` resolves in this order for both UIKit and SwiftUI: (1) `accessibilityLabel`, (2) `accessibilityIdentifier`, (3) `ClassName_<hash>` fallback. Consistent across frameworks.

Sources/Autocapture/SemanticExtractor.swift
rahul-mixpanel: Addressed — `findInteractiveAncestor` now uses `AutocaptureDefaults.maxHierarchyDepth` instead of a hardcoded value.

Sources/Autocapture/DeadClickDetector.swift
rahul-mixpanel: Addressed — DeadClickDetector now uses a cancellable `Task` (iOS 13+) that is cancelled when a new click arrives. Falls back to `DispatchQueue.main.asyncAfter` on older iOS.

Sources/Autocapture/AutocaptureOptions.swift
rahul-mixpanel: Addressed — `ClickOptions` doc comment now describes that a `$mp_click` event is tracked for every tap, including the properties captured: touch coordinates, element identifier, class name, accessibility label, semantic role, and view hierarchy path.

Sources/Autocapture/AutocaptureOptions.swift
rahul-mixpanel: Addressed — renamed to `timeWindowMs` for consistency with `RageClickOptions`.

Sources/Autocapture/Model/ClickEvent.swift
rahul-mixpanel: Addressed — both `tagName` and `elements` are now `String?` and default to `nil`.

Sources/Autocapture.swift
rahul-mixpanel: Addressed — redundant code from `trackScreenView` and `trackScreenLeave` has been moved into a shared `trackScreenEvent` helper, which routes through `trackAutocaptureEvent`.

Mixpanel-swift.podspec
rahul-mixpanel: Addressed — podspec now uses `Sources/**/*.swift` to include all Swift files from the Sources folder.

Sources/MixpanelInstance.swift
rahul-mixpanel: Addressed — the `Autocapture` object is created early, but the `AutocaptureManager` (which starts actual capture) is initialized after the `optOutTrackingByDefault` check. If the user has opted out, autocapture won't start.

Sources/MixpanelOptions.swift
rahul-mixpanel: There are only two inits, separated by `#if os(iOS)` for platform-specific parameters (`autocaptureOptions` is iOS-only). This is the standard pattern for conditional compilation — not init proliferation.

Sources/Autocapture/SemanticExtractor.swift
```diff
@@ -0,0 +1,214 @@
...
+    /// 3. `ClassName_<hash>`
+    private func generateElementId(for view: UIView) -> String {
+      // accessibilityLabel is primary for both UIKit and SwiftUI
+      if let label = findAccessibilityLabel(in: view), !label.isEmpty {
+        return label
+      }
+      // accessibilityIdentifier as fallback
+      if let identifier = findAccessibilityIdentifier(in: view), !identifier.isEmpty {
+        return identifier
+      }
```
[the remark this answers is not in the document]
rahul-mixpanel: this is the design. we don't track the text automatically. however, the dev can supply useful accessibilityLabel to detect which element was clicked.

Sources/Autocapture/AutocaptureOptions.swift
rahul-mixpanel: for now, we have designed it to enable the all at once. you can disable the autocapture options that you don't want. else going forward, the dev will need to add a lot of boilerplate code to enable each autocapture options.

Sources/MixpanelOptions.swift
rahul-mixpanel: This will be addressed in a new PR

context/AUTOCAPTURE.md
rahul-mixpanel: this won't happen as the screen changes when you click back button.

Sources/Autocapture/Model/ClickEvent.swift
```diff
@@ -0,0 +1,155 @@
...
+    /// Use `String(describing: type(of: view))` to get the class name.
+    /// Set to `nil` if not available.
+    public let tagName: String?
+
+    /// The human-readable accessibility label of the element.
+    ///
+    /// This is the text read aloud by VoiceOver — typically the view's `accessibilityLabel`.
+    /// Examples: `"Add to cart"`, `"Play video"`, `"Close"`.
+    /// Set to `nil` if the element has no accessibility label.
+    public let accessibleLabel: String?
```
ketanmixpanel, line 56: Can we not use `accessibilityLabel` instead of `accessibleLabel`  here?

Sources/Autocapture/RageClickTracker.swift
```diff
@@ -0,0 +1,128 @@
...
+
+        // Add current click to history
+        recentClicks.append(ClickRecord(x: x, y: y, timestamp: now))
+
+        // Rage click triggers when we have threshold-1 previous clicks nearby
+        // (so the current click is the Nth click)
+        let isRageClick = nearbyCount >= (clickThreshold - 1)
+
+        // Clean old clicks to prevent memory growth
+        if recentClicks.count > 20 {
```
ketanmixpanel, line 96: lets use a constant instead of hardcoding it to 20

Sources/Autocapture/RageClickTracker.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    // MARK: - Public API
+
+    /// Track a click and check if it triggers a rage click.
+    ///
+    /// - Parameters:
+    ///   - x: X coordinate of the click
+    ///   - y: Y coordinate of the click
+    /// - Returns: Result indicating if this is a rage click and the total tap count
+    func trackClick(x: CGFloat, y: CGFloat) -> RageClickResult {
```
ketanmixpanel, line 76: The `trackClick` function is executed on the main thread. Main thread needs to handoff the work to a different/SDK thread so it does not gets blocked by the SDK processing work.

ketanmixpanel: UI related work is not involved here so the computation can be done in the background.

Sources/Autocapture/DeadClickDetector.swift
```diff
@@ -0,0 +1,284 @@
...
+        )
+        lock.unlock()
+
+        // Schedule final check as a cancellable task
+        let timeWindow = timeWindowMs
+        if #available(iOS 13.0, *) {
+            checkTask = Task { [weak self] in
+                try? await Task.sleep(nanoseconds: UInt64(timeWindow) * 1_000_000)
+                guard !Task.isCancelled else { return }
+                await MainActor.run {
```
ketanmixpanel, line 144: Xcode is giving warning as - 

Reference to captured var 'self' in concurrently-executing code; this is an error in the Swift 6 language mode

```suggestion
          await MainActor.run { [self] in
```

Sources/Autocapture/RageClickTracker.swift
rahul-mixpanel: Done in #765 — it is now `AutocaptureDefaults.maxTrackedClicks`, alongside the other autocapture tunables, with a note that it bounds memory only (clicks outside the time window are already ignored when counting, so it never affects detection).

Sources/Autocapture/RageClickTracker.swift
rahul-mixpanel: Done in #765. `AutocaptureManager.processTouch` now keeps only the work that must be on the main thread — `extractSemantics` (reads the view hierarchy) and `DeadClickDetector.startMonitoring` (its baseline window snapshot has to be captured before the app's click handler can change the UI, or a fast response gets absorbed into the baseline and the click looks dead). Rage click bookkeeping and $mp_click/$mp_rage_click emission moved to a serial `com.mixpanel.autocapture.processing` queue; serial so clicks stay in the order the user made them, which rage detection depends on.

`RageClickTracker` keeps its synchronous NSLock-guarded API, so no call-site or test churn.

One consequence: emission is no longer inline with the touch, so a test that read the event queue straight after a tap (`testClickEventHasTokenProperty`) now waits via `waitForEvent`.

Sources/Autocapture/DeadClickDetector.swift
rahul-mixpanel: Done in #765, with `guard let self else { return }` before the hop so the closure captures an immutable binding rather than the captured weak var. Verified with `SWIFT_STRICT_CONCURRENCY=complete`: "reference to captured var 'self' in concurrently-executing code" no longer appears.

Other Swift 6 warnings remain in this file (non-Sendable `DeadClickDetector` captured in @Sendable closures, main-actor isolation) — those need the type to be Sendable or @MainActor, which felt like its own change rather than a drive-by. Happy to pick it up separately if you want the module strict-concurrency clean.

Sources/Autocapture/Model/ClickEvent.swift
rahul-mixpanel: Left as `accessibleLabel` for now — it is named for parity with Android's `ClickEvent.accessibleLabel`, and it serialises to `$attr-aria-label` on both platforms, so the payload is identical either way. Your point about matching UIKit's spelling is fair though; if we want it, the cleaner move is renaming both SDKs together rather than letting them drift. Happy to do that as a pair if you would prefer it.

Sources/MixpanelInstance.swift
```diff
@@ -550,6 +555,27 @@ open class MixpanelInstance: CustomDebugStringConvertible, FlushDelegate, AEDele
             optOutTracking()
         }
 
+        // Initialize autocapture if enabled (iOS only)
+        // Done after opt-out check so autocapture is not started when tracking is opted out.
+        // Note: optOutTracking() runs async on trackingQueue, so hasOptedOutTracking() may
+        // not reflect the default yet. Check optOutTrackingByDefault directly as well.
+        #if os(iOS)
```
ketanmixpanel, line 562: An additional check for the  mac catalyst is required here. with mac catalyst you can run the iphone/ipad app on the mac. So if this functionality needs to be blocked for the mac catalyst, then a additional check is required here. 
```
        #if os(iOS) && !targetEnvironment(macCatalyst)

```

ketanmixpanel (approved):

PR #758, opened 2026-08-03
feat: walk-up-to-clickable-parent for autocapture

rahul-mixpanel:
## Summary
- When a non-interactive leaf view is tapped, walk up the view hierarchy to the nearest interactive ancestor and use its identity for `$el_id`
- Walk-up is always-on and uses interactivity checks (UIControl with targets, UITapGestureRecognizer, `.button` trait)
- Add walk-up demo test screens for both UIKit and SwiftUI
- Add instrumented tests validating walk-up behavior for UIKit views

## Test plan
- [x] Walk-up resolves to interactive parent's `$el_id` for non-interactive leaves
- [x] Non-interactive leaf with own `accessibilityLabel` still walks up (leaf identity ignored)
- [x] Interactive leaf (UIButton) does NOT walk up (owns its own click)
- [x] Nested interactives stop at the innermost interactive parent
- [x] Deep nesting (9 levels) finds interactive ancestor within 10-level limit
- [x] Non-interactive leaf with no interactive ancestor falls back to hash
- [ ] Run Xcode tests on simulator

### DIFF
--- a/Sources/Autocapture/SemanticExtractor.swift
+++ b/Sources/Autocapture/SemanticExtractor.swift
@@ -7,18 +7,18 @@
 //
 
 #if os(iOS)
-  import UIKit
+import UIKit
 
-  /// Extracts semantic information from UIKit and SwiftUI views for autocapture events.
-  ///
-  /// Handles element identification and role detection.
-  final class SemanticExtractor {
+/// Extracts semantic information from UIKit and SwiftUI views for autocapture events.
+///
+/// Handles element identification and role detection.
+final class SemanticExtractor {
     // MARK: - Configuration
 
     private let autocaptureOptions: AutocaptureOptions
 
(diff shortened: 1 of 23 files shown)


Sources/Autocapture/SemanticExtractor.swift
```diff
@@ -13,6 +13,14 @@ import UIKit
 ///
 /// Handles element identification and role detection.
 final class SemanticExtractor {
+    // MARK: - Configuration
+
+    private let autocaptureOptions: AutocaptureOptions
```
ketanmixpanel, line 18: what is the purpose of passing the `autocaptureOptions` here? I could not find any reference of using it.

ketanmixpanel: @rahul-mixpanel can you confirm this?

ketanmixpanel (approved):

PR #763, opened 2026-08-17
feat: Implement semver and date custom operators for flags runtime events

efahk:
Adds Semantic Versioning and RFC3339 date comparison support for feature flags runtime event matching. This implements them as extensions to the JSONLogic library that is being used for targeting.

See [here](https://app.notion.com/p/mxpnl/Feature-Flag-Custom-Operators-3c0e0ba92562806a9466d04c219f171f) for more information about the comparators and golden vector list


Package.swift
```diff
@@ -29,6 +29,7 @@ let package = Package(
             dependencies: [
                 .product(name: "MixpanelSwiftCommon", package: "mixpanel-swift-common"),
                 .product(name: "jsonlogic", package: "json-logic-swift"),
+                .product(name: "JSON", package: "json-logic-swift"),
```
tylerjroach, line 32: Just wanting a quick check that this additional dependencies inclusion (we already have other json-logic-swift imports) is needed. Ok if, so, but wanted to check and see if it were.

efahk: Yep, the JSON module inclusion is explicitly needed as part of the custom op implementation

tylerjroach (approved):

PR #764, opened 2026-08-19
feat: add pluggable ElementIdExtractor for $el_id resolution

rahul-mixpanel:
## Summary

Adds a public `ElementIdExtractor` protocol so host apps decide exactly which identifier autocapture reports as `$el_id` — the main lever for keeping PII out of autocapture payloads. With no extractor supplied, an internal `DefaultElementIdExtractor` resolves it.

Companion PR: the same API on Android (`mixpanel-android`, branch `feature/custom-element-id-extractor`).

## Resolution order

**UIKit / React Native**

1. React Native `nativeID` — read through the Objective-C runtime (`responds(to:)` then `value(forKey:)`), so the SDK carries no compile-time React Native dependency
2. `accessibilityIdentifier` — stable, developer-assigned, never user-visible; framework-internal identifiers (`_…`, `AXID-…`) are skipped
3. `accessibilityLabel` — only when intentionally set. Known UIKit controls and SwiftUI views are trusted; a generic container (e.g. React Native's `RCTView`) only when `isAccessibilityElement` is true, which maps to `accessible={true}` — so UIKit's auto-derived container labels cannot leak
4. `<ClassName>_<hash>`

**SwiftUI** collapses to `accessibilityIdentifier` > `accessibilityLabel` > hash. The `nativeID` probe is skipped, since React Native renders through UIKit and never SwiftUI. Neither `.accessibilityIdentifier(…)` nor `.accessibilityLabel(…)` reaches the backing `UIView`, so `SemanticExtractor` resolves both from SwiftUI's accessibility element tree — the new `findSwiftUIAccessibilityIdentifier(at:view:)` mirrors the existing label helper — and passes them in as fallback arguments, each consulted only after the view's own property for that step.

## Behavior changes

- `accessibilityIdentifier` now outranks `accessibilityLabel` (previously reversed). The label is unaffected as `$attr-aria-label` — only `$el_id` changed.
- A custom extractor is authoritative: a `nil`/empty return yields the anonymous `<ClassName>_<hash>` rather than falling back to view metadata the developer declined to expose.

`AutocaptureOptions.elementIdExtractor` is declared under `#if os(iOS)` with a split init, so tvOS/macOS/watchOS still compile (`swift build` verified).

## Testing

78 tests pass on an iPhone 17 simulator:

- `DefaultElementIdExtractorTests` (18) — each priority step, the priority pairs, the internal-identifier skip, the container label guard in both directions, empty strings, the SwiftUI fallback parameters, hash shape/stability. Uses a `FakeReactNativeView` with an `@objc nativeID` so `responds(to:)`/`value(forKey:)` behave exactly as on a real `RCTView`.
- `SemanticExtractorElementIdTests` (5) — the options-to-`ClickEvent` wiring, including that a nil-returning extractor does not leak the suppressed identifier.
- `ElementIdExtractorEndToEndTests` (4) — real touches through `Mixpanel.initialize`, covering a React Native `nativeID` view end to end.
- One new test each in the UIKit and SwiftUI instrumented suites pinning identifier-over-label on a real tap.

Also verified in a real app: `mixpanel-react-native`'s `MixpanelStarter` sample, built against this checkout as a local pod, reports `"$el_id":"checkout_button"` for a button with `nativeID="checkout_button"` while `$attr-aria-label` still carries its accessibility label.

## Notes for reviewers

- The two SwiftUI accessibility-tree helpers cannot be covered by tests: SwiftUI only materializes that tree while an accessibility client is attached, so `accessibilityHitTest` returns nil in an XCTest host. Verified against the leaf view, `HostingScrollView`, `_UIHostingView` and the window, in both coordinate spaces. The caveat is documented on the helper; the existing label helper has the same constraint.
- The first commit on this branch is pre-existing WIP (accessibilityLabel auto-derivation verification tests plus `SceneDelegate.swift`) that was carried over from the base branch and had to be restored for the demo app to build. It is committed separately so this feature stays reviewable on its own.
- `xcodebuild test` on the demo project needs `IPHONEOS_DEPLOYMENT_TARGET=15.0`; without it the test target builds for iOS 12 against the iOS 15 app module and fails. Pre-existing project config, not addressed here.

### DIFF
--- a/Sources/Autocapture/AutocaptureOptions.swift
+++ b/Sources/Autocapture/AutocaptureOptions.swift
@@ -211,11 +211,36 @@ public struct AutocaptureOptions {
     /// Configuration for dead click detection.
     public let deadClickOptions: DeadClickOptions
 
+    #if os(iOS)
+    /// Optional custom resolver for the `$el_id` reported on autocaptured interactions.
+    ///
+    /// When `nil` (the default), the SDK resolves `$el_id` internally: React Native `nativeID`,
+    /// then `accessibilityIdentifier`, then `accessibilityLabel`, then an anonymous `<ClassName>_<hash>`
+    /// identifier. Supply an implementation to control exactly which identifier is reported and to
+    /// keep personally identifiable information out of the payload.
+    ///
+    /// - SeeAlso: ``ElementIdExtractor``
+    public let elementIdExtractor: ElementIdExtractor?
+    #endif
+
     /// Returns `true` if any autocapture feature is enabled.
     public var isEnabled: Bool {
         return clickOptions.enabled || rageClickOptions.enabled || deadClickOptions.enabled
     }
 
+    #if os(iOS)
+    public init(
+        clickOptions: ClickOptions = ClickOptions(),
+        rageClickOptions: RageClickOptions = RageClickOptions(),
+        deadClickOptions: DeadClickOptions = DeadClickOptions(),
+        elementIdExtractor: ElementIdExtractor? = nil
+    ) {
+        self.clickOptions = clickOptions
+        self.rageClickOptions = rageClickOptions
+        self.deadClickOptions = deadClickOptions
+        self.elementIdExtractor = elementIdExtractor
+    }
+    #else
     public init(
         clickOptions: ClickOptions = ClickOptions(),
         rageClickOptions: RageClickOptions = RageClickOptions(),
@@ -225,4 +250,5 @@ public struct AutocaptureOptions {
         self.rageClickOptions = rageClickOptions
         self.deadClickOptions = deadClickOptions
     }
+    #endif
 }
--- a/Sources/Autocapture/ElementIdExtractor.swift
+++ b/Sources/Autocapture/ElementIdExtractor.swift
@@ -1,0 +1,194 @@
+//
+//  ElementIdExtractor.swift
+//  Mixpanel
+//
+//  Copyright (c) Mixpanel. All rights reserved.
+//
+
+#if os(iOS)
+import UIKit
+
+// MARK: - ElementIdExtractor
+
+/// Resolves the `$el_id` property reported for an autocaptured interaction.
+///
+/// Provide an implementation via `AutocaptureOptions(elementIdExtractor:)` to take full control
+/// over which identifier the SDK reports for a tapped view. This is the recommended way to keep
+/// personally identifiable information out of autocapture events: the SDK only ever reports what
+/// this method returns.
+///
+/// When no extractor is provided, the SDK falls back to an internal default implementation that
+/// resolves the identifier from the React Native `nativeID`, then `accessibilityIdentifier`, then
+/// `accessibilityLabel`.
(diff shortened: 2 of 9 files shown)


Sources/Autocapture/AutocaptureOptions.swift
```diff
@@ -211,11 +211,36 @@ public struct AutocaptureOptions {
...
+    /// - SeeAlso: ``ElementIdExtractor``
+    public let elementIdExtractor: ElementIdExtractor?
+    #endif
+
     /// Returns `true` if any autocapture feature is enabled.
     public var isEnabled: Bool {
         return clickOptions.enabled || rageClickOptions.enabled || deadClickOptions.enabled
     }
 
+    #if os(iOS)
```
ketanmixpanel, line 231: The autocapture is iOS only right? then why do we need separate inits here?

Sources/Autocapture/ElementIdExtractor.swift
```diff
@@ -0,0 +1,194 @@
...
+    /// Resolution with optional precomputed values used at priority steps 2 and 3.
+    ///
+    /// SwiftUI renders its views as internal UIKit views that carry neither
+    /// `accessibilityIdentifier` nor `accessibilityLabel` on the UIKit view itself; the caller
+    /// resolves those from SwiftUI's accessibility element tree and passes them here so SwiftUI
+    /// elements still get a meaningful `$el_id`. Each fallback is consulted only after the view's
+    /// own property for that step, so an identifier always outranks any label.
+    func elementId(
+        for view: UIView,
+        accessibilityIdentifierFallback: String?,
```
ketanmixpanel, line 96: what is use of the `accessibilityIdentifierFallback`?

Sources/Autocapture/ElementIdExtractor.swift
```diff
@@ -0,0 +1,194 @@
...
+    /// `RCTView` (and friends) expose `nativeID` as an Objective-C property carrying the JS-side
+    /// `nativeID` prop. `responds(to:)` is checked first so KVC never throws on views that do not
+    /// declare the key.
+    ///
+    /// Skipped entirely for SwiftUI views — React Native renders through the UIKit view hierarchy,
+    /// so a SwiftUI view can never carry a `nativeID`.
+    private func reactNativeId(for view: UIView) -> String? {
+        guard !AutocaptureDefaults.isSwiftUIView(view) else { return nil }
+        let selector = NSSelectorFromString("nativeID")
+        guard view.responds(to: selector) else { return nil }
```
ketanmixpanel, line 143: is this check really required? if `view.value` is what we are doing what is purpose of `view.responds(to: selector)`?

ketanmixpanel (approved):

Sources/Autocapture/AutocaptureOptions.swift
rahul-mixpanel: we will remove this. for now, we'll keep only the default extractor and won't accept one from the host app.

PR #766, opened 2026-08-20
fix: address autocapture review feedback on threading and constants

rahul-mixpanel:
Review fixes for #746, kept on their own branch so #746's diff stays about the feature.

Move non-UI touch processing off the main thread (RageClickTracker.swift comments)

  Touches arrive on the main thread and AutocaptureManager did all of its work there.
  Only part of it has to be: extractSemantics reads the view hierarchy, and
  DeadClickDetector.startMonitoring must capture its baseline window snapshot
  synchronously at click time — otherwise a fast click handler updates the UI first,
  the response is absorbed into the baseline, and the click looks dead.

  Everything after that is pure computation over already-extracted values, so rage
  click bookkeeping and $mp_click/$mp_rage_click emission now run on a serial
  com.mixpanel.autocapture.processing queue. Serial so clicks stay in the order the
  user made them, which rage click detection depends on.

  RageClickTracker keeps its synchronous, NSLock-guarded API — no call-site or test
  churn, and it was already safe to call from another thread.

  Consequence worth knowing: click emission is no longer inline with the touch, so
  the event's timestamp moves by roughly a millisecond and anything reading the event
  queue immediately after a tap must wait. testClickEventHasTokenProperty did exactly
  that and now waits for the event via waitForEvent, which polls and drains the
  tracking queue; its assertions are unchanged.

Extract the rage click history bound (RageClickTracker.swift:96)

  The hardcoded 20 becomes AutocaptureDefaults.maxTrackedClicks, alongside the other
  autocapture tunables, with a note on what it does and does not affect: it bounds
  memory only, since clicks outside the time window are already ignored when counting.

Fix the Swift 6 capture warning (DeadClickDetector.swift:144)

  Bind self before hopping to the main actor so the concurrently-executing closure
  captures an immutable binding rather than the captured weak var. Verified with
  SWIFT_STRICT_CONCURRENCY=complete: "reference to captured var 'self' in
  concurrently-executing code" is gone.

  ClickEvent is declared Sendable — all its stored properties are value types — so
  handing it from the main thread to the processing queue introduces no new
  concurrency warnings.

35 autocapture tests pass on an iPhone 17 simulator (AutocaptureUIKitInstrumentedTests, AutocaptureSwiftUIInstrumentedTests, RageClickTrackerTests, AutocaptureOptionsTests, ClickEventTests).

### DIFF
--- a/Sources/Autocapture/DeadClickDetector.swift
+++ b/Sources/Autocapture/DeadClickDetector.swift
@@ -141,8 +141,12 @@ final class DeadClickDetector {
             checkTask = Task { [weak self] in
                 try? await Task.sleep(nanoseconds: UInt64(timeWindow) * 1_000_000)
                 guard !Task.isCancelled else { return }
-                await MainActor.run {
-                    self?.performFinalCheck()
+                // Bind before hopping to the main actor: referencing the captured `weak var`
+                // from inside the concurrently-executing closure is a warning under Swift 5 and
+                // an error in the Swift 6 language mode.
+                guard let self = self else { return }
+                await MainActor.run { [self] in
+                    self.performFinalCheck()
                 }
             }
         } else {
--- a/MixpanelDemo/MixpanelDemoTests/AutocaptureUIKitInstrumentedTests.swift
+++ b/MixpanelDemo/MixpanelDemoTests/AutocaptureUIKitInstrumentedTests.swift
@@ -434,20 +434,14 @@ class AutocaptureUIKitInstrumentedTests: MixpanelBaseTests {
         // When: Simulate tap
         simulateTap(on: button)
 
-        // Wait and check event queue for full properties
-        waitForTrackingQueue(mixpanel)
+        // Wait for the event rather than reading the queue straight after the tap: autocapture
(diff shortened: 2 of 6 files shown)


Sources/Autocapture/DeadClickDetector.swift
```diff
@@ -141,8 +141,12 @@ final class DeadClickDetector {
             checkTask = Task { [weak self] in
                 try? await Task.sleep(nanoseconds: UInt64(timeWindow) * 1_000_000)
                 guard !Task.isCancelled else { return }
-                await MainActor.run {
-                    self?.performFinalCheck()
+                // Bind before hopping to the main actor: referencing the captured `weak var`
+                // from inside the concurrently-executing closure is a warning under Swift 5 and
+                // an error in the Swift 6 language mode.
+                guard let self = self else { return }
+                await MainActor.run { [self] in
```
ketanmixpanel, line 148: the injected self has to be weak self. instead you can pass the existing weak self as is to the `MainActor.run`.

ketanmixpanel (changes requested):

PR #767, opened 2026-08-20
fix: never use accessibilityLabel as $el_id, and drop $attr-aria-label

rahul-mixpanel:
## Why

`accessibilityLabel` is user-facing text. Two problems with treating it as identity:

- **It is localized.** The same button reports `Checkout` in English and `Pagar` in Spanish, so a single element splits into one `$el_id` per language and cannot be grouped.
- **It can carry personal data.**

The same reasoning applies to reporting it at all, so `$attr-aria-label` is removed from the payload.

Stacked on #766 — merges into `fix/review-comments`. Companion Android PR: mixpanel/mixpanel-android#997.

## What changed

Resolution is now **nativeID → accessibilityIdentifier → `<ClassName>_<hash>`**. The nativeID step is still skipped for SwiftUI, and the identifier is still resolved from SwiftUI's accessibility element tree when the backing `UIView` has none.

Removed: the `accessibilityLabel` step and `intentionalAccessibilityLabel` from `DefaultElementIdExtractor`, the `accessibilityLabelFallback` parameter, `SemanticExtractor.findAccessibilityLabel` and `findSwiftUIAccessibilityLabel`, and `ClickEvent.accessibleLabel` with its `$attr-aria-label` output. The `ClickEvent` change is a public API break, acceptable because autocapture is unreleased.

## The hash had to change too

With the label gone, an element with no identifier falls to `<ClassName>_<hash>` — and that hash was `view.hash`. Swift seeds hashing per process, so the id differed on **every launch**: un-instrumented screens would have produced a fresh `$el_id` each session and never grouped.

The hash now describes **where the element sits**:

```
UIButton@UIStackView[2]/UIScrollView[0]/UIView[1]
```

Class names and sibling indices only — no user text. It is hashed with **FNV-1a over the UTF-8 bytes**, not `hashValue`, precisely because `hashValue` is per-process seeded and would reintroduce the instability this change removes. The payload shape is unchanged: `<ClassName>_<hash>`.

Worth knowing: this identifies a **position, not a row** — reordering siblings changes the id, and two rows of one list differ by index rather than content.

## Tests — 76/76 pass on an iPhone 17 simulator

- The label-priority cases in `DefaultElementIdExtractorTests` are replaced by a test that no view shape — `UILabel`, `UIButton`, plain `UIView` — ever resolves to its label, whatever its `isAccessibilityElement` state.
- The "label is captured" tests in the UIKit and SwiftUI suites became leak-regression tests: an intentional label must appear in **no** property.
- New coverage for the structural hash: identical for the same layout built twice, different for siblings at different positions.
- `swift build` still passes for the non-iOS platforms.

## Before merging

Anything downstream consuming `$attr-aria-label` — dashboards, session-replay correlation — loses that property. Worth a check. Android keeps the same shape, so the two SDKs stay aligned.

### DIFF
--- a/Sources/Autocapture/ElementIdExtractor.swift
+++ b/Sources/Autocapture/ElementIdExtractor.swift
@@ -18,8 +18,9 @@ import UIKit
 /// this method returns.
 ///
 /// When no extractor is provided, the SDK falls back to an internal default implementation that
-/// resolves the identifier from the React Native `nativeID`, then `accessibilityIdentifier`, then
-/// `accessibilityLabel`.
+/// resolves the identifier from the React Native `nativeID`, then `accessibilityIdentifier`, then a
+/// structural fallback. `accessibilityLabel` is deliberately not a source: it is localized, so the
+/// same element would report a different identifier per language, and it can carry user data.
 ///
 /// **Example:**
 /// ```swift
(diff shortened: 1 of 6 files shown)


Sources/Autocapture/ElementIdExtractor.swift
```diff
@@ -184,11 +148,51 @@ final class DefaultElementIdExtractor: ElementIdExtractor {
...
         let className = String(describing: type(of: view))
-        // abs(Int.min) traps, so map it to Int.max.
-        let safeHash = view.hash == Int.min ? Int.max : abs(view.hash)
-        return "\(className)_\(String(safeHash, radix: 16))"
+        return "\(className)_\(stableHash(structuralPath(for: view)))"
+    }
+
+    /// Class names and sibling indices only — never text — so the path cannot carry user data:
+    /// `UIButton@UIStackView[2]/UIScrollView[0]/UIView[1]`.
+    static func structuralPath(for view: UIView) -> String {
```
ketanmixpanel, line 165: I think this structural path is going to be redundant. The elements attribute is also going to have same information?

rahul-mixpanel: Oh, my plan was to append a hash of this to UIButton (eg: UIButton_asdfg). Let me check why this is different here.

ketanmixpanel (approved):

PR #768, opened 2026-08-20
feat: mark autocapture as experimental (beta)

rahul-mixpanel:
## Why

Autocapture works and is safe to try, but it is not at GA: element identification depends on how a screen is built, the API can still change, and captured property shapes can be refined.

Stacked on #767 — merges into `fix/el-id-drop-accessibility-labels`. Companion Android PR: mixpanel/mixpanel-android#999.

## Approach — identical on both platforms

Per team decision, Android and iOS signal beta status the same way:

1. **An "Experimental (beta)" note in the documentation** of every public autocapture symbol, shown in Xcode Quick Help and generated docs.
2. **One warn-level log line when autocapture starts**, for developers who never open the docs:

```
Autocapture is experimental (beta): it may contain issues, and its API and captured properties
may change before general availability. Pin your SDK version if you build reports on
autocaptured events.
```

No compile-time gate on either platform. For the record, the two Swift mechanisms that would warn at the call site were both rejected: `@available(*, deprecated, message:)` tells developers a brand-new feature is *being removed*, and would warn throughout `MixpanelDemo` and both React Native samples; `@_spi(Experimental)` would force `MixpanelOptions`' single shared initializer to become SPI — so every developer, autocapture or not, would need `@_spi(Experimental) import Mixpanel` — besides being invisible to Objective-C and an unsupported underscored attribute.

## Documented surface

`AutocaptureOptions`, `ClickOptions`, `RageClickOptions`, `DeadClickOptions`, `ClickEvent`, the `ElementIdExtractor` protocol, `MixpanelOptions.autocaptureOptions`, `MixpanelInstance.autocapture`, and the `Autocapture` facade.

Drive-by fix: `ClickOptions`' documentation still listed the accessibility label among captured properties — untrue since labels were dropped from capture in #767.

## Removal at GA

Delete the callouts and one log statement. They are comments, so nothing a consumer wrote can break.

## Testing

43 autocapture tests pass on an iPhone 17 simulator; `swift build` is clean for the non-iOS platforms.

### DIFF
--- a/Sources/Autocapture.swift
+++ b/Sources/Autocapture.swift
@@ -11,6 +11,9 @@ import Foundation
 
 /// Access to autocapture tracking methods, available as an accessible variable from
 /// the main Mixpanel instance.
+///
+/// - Warning: **Experimental (beta).** Autocapture may contain issues, and its API and the
+///   properties it captures may change in a future release before general availability.
 open class Autocapture {
 
     weak var mixpanelInstance: MixpanelInstance?
--- a/Sources/Autocapture/Model/ClickEvent.swift
+++ b/Sources/Autocapture/Model/ClickEvent.swift
@@ -17,6 +17,10 @@ import UIKit
 /// All stored properties are value types (`CGFloat`, `String`, `Bool`), so the event can safely
 /// cross threads — it is handed from the main thread, where it is extracted, to the queue that
 /// emits it.
+///
+/// - Warning: **Experimental (beta).** Autocapture may contain issues, and its API and the properties
+///   it captures may change in a future release before general availability. Pin your SDK version if
+///   you build reports on autocaptured events.
 public struct ClickEvent: Sendable {
     // MARK: - Position
 
--- a/Sources/Autocapture/AutocaptureManager.swift
+++ b/Sources/Autocapture/AutocaptureManager.swift
@@ -89,6 +89,13 @@ final class AutocaptureManager {
     // MARK: - Lifecycle
 
     /// Start autocapture by installing the touch interceptor.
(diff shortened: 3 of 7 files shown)


Sources/Autocapture/Model/ClickEvent.swift
```diff
@@ -17,6 +17,10 @@ import UIKit
 /// All stored properties are value types (`CGFloat`, `String`, `Bool`), so the event can safely
 /// cross threads — it is handed from the main thread, where it is extracted, to the queue that
 /// emits it.
+///
+/// - Warning: **Experimental (beta).** Autocapture may contain issues, and its API and the properties
+///   it captures may change in a future release before general availability. Pin your SDK version if
```
ketanmixpanel, line 22: Do you want to mention something like adding dynamic disabling as a way to roll back the feature if you plan to release it to production?

Sources/Autocapture.swift
```diff
@@ -11,6 +11,9 @@ import Foundation
 
 /// Access to autocapture tracking methods, available as an accessible variable from
 /// the main Mixpanel instance.
+///
+/// - Warning: **Experimental (beta).** Autocapture may contain issues, and its API and the
+///   properties it captures may change in a future release before general availability.
 open class Autocapture {
```
ketanmixpanel, line 17: tracking screens is not experimental right?

ketanmixpanel: can you confirm?

ketanmixpanel (approved):
Approving the PR. The comments can be addressed in a separate PR based on their priority.

PR #785, opened 2026-09-10
docs: update README links and clarify SDK options

santigracia:
## Summary
- Fix broken/outdated documentation links, pointing them to `docs.mixpanel.com` (Advanced Swift Guide, user profiles, opt-out, Lexicon blocking data, default properties, legacy automatic events)
- Add a top-of-README pointer to the official Swift SDK docs
- Add a **Data Residency** note to the "Why aren't my events showing up?" section, with an EU endpoint `MixpanelOptions` example
- Document configuring `flushInterval` via `MixpanelOptions` at init time
- Clarify that common mobile events are now legacy and disabled by default (enabled via `trackAutomaticEvents`), while default event/user properties are still tracked
- Update Lexicon wording from "drop" to "block" to match current docs
- Refresh support links to `mixpanel.com/get-support` and add the Support Hub

## Test plan
- [ ] Verify all updated links resolve correctly
- [ ] Confirm `MixpanelOptions` code examples (`serverURL`, `flushInterval`, `trackAutomaticEvents`) match the current SDK API

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -18,13 +18,13 @@ jobs:
         run: |
 
           # Conventional commit pattern
-          # Supports: feat, fix, chore, release
+          # Supports: feat, fix, chore, docs, release
           # Examples:
           #   feat: Add new feature
           #   fix: Resolve bug
           #   release: prepare 6.3.0
 
-          PATTERN='^(feat|fix|chore|release): .+$'
+          PATTERN='^(feat|fix|chore|docs|release): .+$'
 
           if [[ ! "$PR_TITLE" =~ $PATTERN ]]; then
             echo "❌ PR title does not follow conventional commit format"
@@ -39,12 +39,14 @@ jobs:
             echo "  feat    - New feature"
             echo "  fix     - Bug fix"
             echo "  chore   - Maintenance task"
+            echo "  docs    - Documentation update"
             echo "  release - Release preparation"
             echo ""
             echo "Valid examples:"
             echo "  ✅ feat: Add automatic event tracking"
             echo "  ✅ fix: Resolve SwiftUI accent color override"
             echo "  ✅ chore: Update dependencies"
+            echo "  ✅ docs: Update integration guide"
             echo "  ✅ release: prepare 6.3.0"
             echo ""
             echo "Invalid examples:"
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -18,13 +18,13 @@ jobs:
...
           # Conventional commit pattern
-          # Supports: feat, fix, chore, release
+          # Supports: feat, fix, chore, docs, release
           # Examples:
           #   feat: Add new feature
           #   fix: Resolve bug
           #   release: prepare 6.3.0
 
-          PATTERN='^(feat|fix|chore|release): .+$'
+          PATTERN='^(feat|fix|chore|docs|release): .+$'
```
ketanmixpanel, line 27: `chore` can be used for docs update too. I don't see any additional benefit of having separate `docs` tag.

Also these options are added across all the other SDK repositories, adding it to just one will make it inconsistent with other SDK workflows.

santigracia: indeed, reverting to use chore, it was Copilot adding "docs"

README.md
```diff
@@ -162,14 +167,14 @@ func application(_ application: UIApplication,
...
+Let's get started by sending event data. You can send an event from anywhere in your application. Better understand user behavior by storing details that are specific to the event (properties). After initializing the library, Mixpanel can [automatically collect common mobile events](https://docs.mixpanel.com/docs/tracking-methods/sdks/swift#legacy-automatically-tracked-events). These events are now considered legacy and are disabled by default. You can enable them by setting `trackAutomaticEvents` to `true` in `MixpanelOptions` when initializing the library. Regardless of this setting, Mixpanel still automatically tracks some [default event and user properties](https://docs.mixpanel.com/docs/data-structure/property-reference/default-properties#client-side-sdks).
```
ketanmixpanel: as autocapture is now out, can we link the Auto capture docs instead of recommending legacy feature?
https://docs.mixpanel.com/docs/tracking-methods/sdks/swift#autocapture

ketanmixpanel (changes requested):

README.md
santigracia: thanks! good call! 🙏, I'm gonna add autocapture and mention legacy automatic events in the FAQ

PR #786, opened 2026-09-15
fix: Flush queue optimisation

ketanmixpanel:
**Flush queue optimization: batch draining, crash fixes, and reliability improvements**

**Summary**

Reworks the flush pipeline to drain the local queue in bounded batches instead of unbounded reads, fixing several crash and memory-spike conditions (including an uncatchable `NSMallocException` raised inside `JSONSerialization` during `MPDB.readRows`), and closes correctness gaps found during review: infinite-recursion risk, premature drain stop, background-task lifecycle, stale backoff state, and the flush being unobservable to the host app.

**Key changes**

Flush loop (`Flush.swift`, `MixpanelInstance.swift`)
- Drain the queue in `flushBatchSize`-sized batches (also bounded by a byte budget) instead of reading everything into memory at once. Only one batch of one queue type is alive at a time.
- Drain queue types sequentially: events → people → groups, matching the Android SDK. A failed send advances to the next type instead of retrying the same one, so one struggling type cannot consume both pre-backoff retries and starve the others.
- Continue draining while any read returned rows, even a short (non-full) batch — previously a short batch was mistaken for "queue empty" and could strand deliverable events until the next flush cycle.
- Stop chasing a queue type once a batch's last row was captured after the flush started, so continuous tracking cannot keep the events drain spinning and starve people/groups.
- `flush(performFullFlush:)` now defaults to `true`, and the timer flush performs a full drain instead of sending one batch per type per tick. Memory stays bounded by batching.
- Clamp `flushBatchSize` to a minimum of 1, preventing infinite recursion when set to 0.
- Coalesce overlapping `flush()` calls. The background-task lifecycle (`applicationDidEnterBackground`) is now tied to whether flushing is actually idle rather than to whichever call was rejected by coalescing — fixes a case where entering the background could end the background execution allowance while a flush was still genuinely in flight.
- `flush()` returns `Bool` (`@discardableResult`, so existing callers compile unchanged): `true` when the call started a flush, `false` when it was rejected because a flush is already running or tracking is opted out. Rejected calls still fire their completion immediately.
- `isFlushInProgress` is now public, so an app can check whether a flush is running before scheduling one.
- `mixpanelWillFlush` is asked once per flush, before the first batch, as before this PR. Opt-out and request backoff are still re-checked before every batch.
- Setting `flushInterval` no longer triggers an immediate flush. Batches go out on the timer, so a `serverURL` set right after init is respected by the first send.

Persistence (`MPDB.swift`, `MixpanelPersistence.s
- Every read carries a `LIMIT`; the unbounded `Int.max` path is gone.
- Drop and delete oversized rows (over `maxRowBytto deserialize them — previously these couldexhaust memory and crash via an uncatchable `NSMallocException`, and would recur on every later flush.
- Cap the cumulative bytes deserialized per read at least one row is always read so the queue cannot stall.
- Drop and delete rows that fail JSON deserializaem stuck in the queue forever.

Automatic events (`Track.swift`)
- Removed the pre-existing read-time filter that excluded `$ae_` rows from a flush once `trackAutomaticEventsEnabled` was
turned off after they'd already been persisted. TR and, once this PR bounded reads to a fixed batchsize, could also stall the events queue if enough excluded rows piled up at the head of the table. The write-time gate in
`Track.swift` — which refuses to persist a new `$is off — is now the SDK's only gate, matchingAndroid, whose equivalent flag is immutable and only ever gates at write time. This also matches how the rest of the SDK
treats already-persisted data: `optOutTracking()`than discarding it, and only the explicit `reset()` API purges anything.

Networking (`FlushRequest.swift`, `Network.swift`)
- When the client-side wait for a network responsthe real completion fires, record it as a failureimmediately (new `Reason.semaphoreTimeout`) so backoff state does not go stale and allow an immediate retry.

Tests / project config
- Added regression coverage in `MixpanelFlushMemomalformed rows, bounded reads and byte budget,batch-size clamping, short-batch draining, sequential type ordering, no starvation under continuous tracking, failed-send behavior, flush coalescing with return values andground-task protection, and delegate asked once per flush.
- Registered the new test file with the tvOS and usly iOS-only).
- Minor: removed a `$screen_height` assertion from `MixpanelScreenTrackingTests.swift`; it is not required for the screen-view
event.

**Behavior changes to note in release notes**
- `flush()` default `performFullFlush` is now `true`; timer flushes drain fully.
- No flush fires at SDK init when `flushInterval`its for the first timer tick or a background event.
- Overlapping `flush(completion:)` calls are rejected (return `false`) and their completion fires immediately without sending.
- Automatic events already persisted before `tracs turned off are now sent on the next flush instead of being silently withheld. Only automatic events created after the toggle are affected.                              
**Test plan**                                                                                                         - `MixpanelFlushMemoryTests` passing on iOS (18 t the background-task test needs a UIKit backgroundgrant the simulator host lacks) and tvOS. macOS verified via successful test-bundle compile; that scheme does not currrun unit tests (pre-existing, unrelated gap).
- iOS `MixpanelDemoTests`, `MixpanelOptOutTests`, and `MixpanelAutomaticEventsTests` run alongside: 97 tests total, 0 failures.
- Full iOS suite run to confirm no regressions (unrelated pre-existing failures in Autocapture UI-gesture-timing tests and one backup-host test, confirmed present without any o

Fixes SDK-91 SDK-140


Sources/MixpanelInstance.swift
```diff
@@ -1339,13 +1408,59 @@ extension MixpanelInstance {
...
+                //
+                // removeProcessedEntities (called above, per type, on success) enqueues its
+                // delete on trackingQueue asynchronously rather than blocking this thread. That
+                // is safe only because trackingQueue is serial: every delete for this iteration
+                // is enqueued, in this order, before the recursive call below enqueues the next
+                // read, so FIFO guarantees the deletes run first. This breaks if trackingQueue
+                // is ever made concurrent, or if this recursion is reordered before the sends.
+                let mayHaveMore =
+                    !eventQueue.isEmpty || !peopleQueue.isEmpty || !groupsQueue.isEmpty
+                if mayHaveMore {
```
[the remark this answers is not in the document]
ketanmixpanel: Accepted tradeoff: oversized/malformed rows in the current window are deleted immediately. If all three queues return empty, remaining rows wait for the next flush cycle; valid events stay persisted.

Sources/MixpanelInstance.swift
```diff
@@ -1288,16 +1294,58 @@ extension MixpanelInstance {
...
+        // Callers that flush per-event can pile up enough of those payloads to exhaust memory.
+        // They also duplicate work: rows are only deleted after a successful send, so a flush that
+        // starts mid-send re-reads rows already in flight and sends them twice.
+        //
+        // Dropping the overlapping call loses nothing. Queued data lives in SQLite until it is
+        // acknowledged, so whatever this call would have sent is picked up by the next flush.
+        guard !hasOptedOutTracking(), beginFlush() else {
             if let completion = completion {
                 DispatchQueue.main.async(execute: completion)
             }
```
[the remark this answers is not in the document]
ketanmixpanel: fixed

Sources/MPDB.swift
```diff
@@ -256,32 +256,71 @@ class MPDB {
         }
     }
 
-    func readRows(_ persistenceType: PersistenceType, numRows: Int, flag: Bool = false)
+    func readRows(
+        _ persistenceType: PersistenceType, numRows: Int, flag: Bool = false,
+        byteBudget: Int = APIConstants.maxReadBatchByteSize
+    )
```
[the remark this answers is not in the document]
ketanmixpanel: updated

Sources/MixpanelInstance.swift
```diff
@@ -1288,16 +1294,58 @@ extension MixpanelInstance {
...
+    }
+
+    /// One iteration of the drain loop: read one batch per queue on `trackingQueue`, send it on
+    /// `networkQueue`, then either re-arm for the next batch or release the flush slot.
+    ///
+    /// The loop continues only while every send succeeded AND at least one queue produced a
+    /// full batch. A full batch that was sent has also been deleted, so each continuation is
+    /// guaranteed progress — the loop cannot spin on rows it never removes (for example, rows
+    /// held back by the automatic-events filter or truncated by the byte budget; those end the
+    /// drain early and wait for the next flush). A short batch means that queue is empty.
```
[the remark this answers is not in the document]
ketanmixpanel: updated comments

Sources/FlushRequest.swift
```diff
@@ -70,7 +70,20 @@ class FlushRequest: Network {
...
+            // and will update these same counters then), but that's too late for the caller,
+            // who is about to decide whether to retry based on requestNotAllowed() right now.
+            // Treat the timeout itself as a failure so that check reflects it immediately,
+            // rather than leaving stale state that lets an immediate retry through.
+            MixpanelLogger.warn(
+                message: "Request to \(resource.path) timed out waiting for a response")
+            self.handleFlushFailure(
+                path: resource.path, reason: .semaphoreTimeout, response: nil, completion:  { success in
+                    result = success
+                })
```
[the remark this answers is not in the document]
ketanmixpanel: Accepted tradeoff: URLSession’s default 60-second timeout handles stalled requests before our 120-second wait, but it is an inactivity timeout—not a total-duration limit—so late callbacks remain possible. The 120-second timeout records a failure to enable backoff; a late callback may update retry state again, affecting retry timing. Timeout alone does not delete queued events.

Sources/MixpanelInstance.swift
```diff
@@ -1339,13 +1417,80 @@ extension MixpanelInstance {
...
+        DispatchQueue.main.async { [weak self] in
+            #if !os(OSX) && !os(watchOS)
+            if let self = self {
+                var flushIsActive = false
+                self.flushStateLock.read {
+                    flushIsActive = self.isFlushing
+                }
+
+                if !flushIsActive {
+                    self.endBackgroundTask()
```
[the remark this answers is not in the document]
ketanmixpanel: Accepted limitation: a new flush can start after background execution ends anyway. Moving cleanup or making the idle check atomic cannot eliminate that broader case; protecting every background flush requires acquiring background execution time when it starts. Unacknowledged events remain persisted for a later flush.

Sources/Flush.swift
```diff
@@ -140,76 +139,54 @@ class Flush: AppLifecycle {
...
+            #if os(iOS)
+            if !MixpanelInstance.isiOSAppExtension() {
+                delegate?.updateNetworkActivityIndicator(true)
+            }
+            #endif  // os(iOS)
+            let success = flushRequest.sendRequest(
+                requestData,
+                type: type,
+                useIP: useIPAddressForGeoLocation,
+                headers: headers,
```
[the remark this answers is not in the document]
ketanmixpanel: Intentional behavior: failed batches remain persisted and are retried within the same flush until the failure counter activates backoff. Semaphore timeouts now record a failure too, addressing the previously uncounted-timeout retry loop. We accept these immediate retries rather than stopping on the first failure.

ketanmixpanel:
@greptileai update the confidance score

Sources/Flush.swift
```diff
@@ -76,7 +76,6 @@ class Flush: AppLifecycle {
                     _flushInterval = newValue
                 })
 
-            delegate?.flush(performFullFlush: false, completion: nil)
```
ketanmixpanel, line 79: This is removed intentionally to fix the issue of setting the server url post sdk init sending a few events to default US endpoint. Now the batches are flushed based on the interval and not immediately after sdk init. The SDK will accept the server url post init and then based on the flush interval, it will send the events to the new endpoint.

Sources/MixpanelInstance.swift
```diff
@@ -1288,16 +1316,127 @@ extension MixpanelInstance {
...
+    /// different key path. Rows are read `ORDER BY time`, so the batch's last element is its
+    /// most recent — no need to scan the rest.
+    private func lastRowTimestamp(_ queue: Queue, queueState: FlushQueueState) -> Double? {
+        guard let lastRow = queue.last else {
+            return nil
+        }
+        let rawValue: Any?
+        switch queueState {
+            case .events:
+                rawValue = (lastRow["properties"] as? [String: Any])?["time"]
```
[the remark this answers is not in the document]
ketanmixpanel: The time key is part of the SDKs `reservedPropertyKeys` which makes sure to not override it from the super or custom properties of the event.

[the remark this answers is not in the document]
ketanmixpanel: accepting this as a known limitation, no SDK change. The app already has full control over these values by construction (it's the caller of track()), and enforcing non-override for time alone would be an inconsistent half-fix, `distinct_id` and `token` would remain just as overridable. We're not carving out time as a special case.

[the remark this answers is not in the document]
ketanmixpanel: @greptileai reassess the confidence score

MixpanelDemo/MixpanelDemoTests/MixpanelScreenTrackingTests.swift
```diff
@@ -28,7 +28,6 @@ class MixpanelScreenTrackingTests: MixpanelBaseTests {
         XCTAssertEqual(props?["current_page_title"] as? String, "HomeScreen")
         XCTAssertEqual(props?["$mp_autocapture"] as? Bool, true)
         XCTAssertEqual(props?["extra_prop"] as? String, "extra_value")
```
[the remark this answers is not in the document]
ketanmixpanel: removed scree height check as it is not required to be checked for the screen view event.

MixpanelDemo/MixpanelDemoMacTests/MixpanelScreenTrackingTests.swift
```diff
@@ -28,9 +28,8 @@ class MixpanelScreenTrackingTests: MixpanelBaseTests {
         XCTAssertEqual(props?["current_page_title"] as? String, "HomeScreen")
         XCTAssertEqual(props?["$mp_autocapture"] as? Bool, true)
         XCTAssertEqual(props?["extra_prop"] as? String, "extra_value")
-        XCTAssertNotNil(props?["$screen_height"])
```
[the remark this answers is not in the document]
ketanmixpanel: removed scree height check as it is not required to be checked for the screen view event.

MixpanelDemo/MixpanelDemoTests/MixpanelFlushMemoryTests.swift
```diff
@@ -0,0 +1,787 @@
...
+        testMixpanel.delegate = delegate
+        let total = APIConstants.maxBatchSize * 2 + 20
+        for index in 0..<total {
+            testMixpanel.track(event: "event\(index)")
+        }
+        waitForTrackingQueue(testMixpanel)
+
+        let finished = expectation(description: "flush finished")
+        XCTAssertTrue(testMixpanel.flush { finished.fulfill() })
+        wait(for: [finished], timeout: 10)
```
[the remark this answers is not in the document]
ketanmixpanel: the test is to check if DelegateIsAskedOncePerFlush, and it does not make api call in this case. So 10 seconds wait is sufficient.

Sources/MixpanelInstance.swift
```diff
@@ -1288,16 +1333,145 @@ extension MixpanelInstance {
...
```
rahul-mixpanel: **[P1] Use a storage-owned flush watermark.** `properties["time"]` is application-overridable: `Track.track` assigns the SDK timestamp and then merges super/event properties afterward. A queued event with a future custom `time` makes a full flush stop after its first batch; conversely, events added during the flush with backdated timestamps can keep this queue draining and starve people/groups. Millisecond rounding also leaves a same-timestamp boundary race. Please snapshot a database-owned watermark (preferably the maximum SQLite row ID per table at flush start) and restrict the drain to that watermark instead of deriving lifecycle state from payload data. Add tests for future custom event time, backdated events created during a flush, and the same-millisecond boundary.

Sources/MPDB.swift
```diff
@@ -256,32 +256,86 @@ class MPDB {
...
-    func readRows(_ persistenceType: PersistenceType, numRows: Int, flag: Bool = false)
+    /// Reads up to `numRows` rows from the table for `persistenceType`, oldest first, and returns
+    /// them deserialized with their row `id` attached.
+    ///
+    /// The read is bounded in bytes as well as rows: it stops before the row that would push the
+    /// cumulative blob size past `byteBudget`, though the first row is always read so a queue can
+    /// never stall. Rows that cannot be sent are dropped and deleted in the same pass — those over
+    /// `APIConstants.maxRowByteSize` and those that fail JSON deserialization. Dropped rows still
+    /// consume the SQL row limit; only successfully decoded rows contribute to the byte counter.
+    /// An entirely dropped window returns empty even if later rows remain queued.
```
rahul-mixpanel: **[P2] Do not treat an entirely discarded window as end-of-queue.** If all physical rows in the `LIMIT` window are malformed/oversized, this method deletes them but returns `[]`. `flushBatches` interprets that as queue exhaustion and advances, leaving valid rows immediately behind them until a later flush. That is especially risky for the last background flush before suspension or termination. Please return scan/progress metadata (`scannedRowCount`, `droppedRowCount`, or `hasMore`) or continue scanning internally after deleting a fully invalid window, while ensuring every loop makes progress. Add an integration test proving one `flush()` drains valid rows behind a fully malformed window.

Sources/FlushRequest.swift
```diff
@@ -70,7 +70,21 @@ class FlushRequest: Network {
                 result = success
                 semaphore.signal()
             })
-        _ = semaphore.wait(timeout: .now() + 120.0)
+        if semaphore.wait(timeout: .now() + 120.0) == .timedOut {
```
rahul-mixpanel: **[P2] Finalize a timed-out request exactly once.** This records a synthetic failure, but the original URLSession callback remains active and later calls `handleFlushSuccess`/`handleFlushFailure` again. A late success can erase the timeout backoff; a late failure counts the same request twice; and those callbacks can race later reads of the retry state. The batch also remains queued although the original request may have reached the server. Please return/retain the `URLSessionTask`, cancel it on timeout, and protect completion with a synchronized one-shot state so late callbacks cannot mutate retry state. Add late-success and late-failure timeout tests.

Sources/MPDB.swift
```diff
@@ -256,32 +256,86 @@ class MPDB {
...
+                            } else {
+                                MixpanelLogger.warn(
+                                    message:
+                                        "Dropping malformed row \(id) from table \(tableName): JSON deserialization failed"
+                                )
+                                droppedIds.append(id)
                             }
-                            rowsRead += 1
                         } else {
                             logSqlError(message: "No blob found in data column for row in \(tableName)")
```
rahul-mixpanel: **[P2] Delete NULL/zero-length blobs as malformed rows.** `sqlite3_column_blob` can return nil for NULL or zero-length data, but this branch only logs and leaves the row in SQLite. Because the schema permits NULL, enough such rows can occupy every bounded read forever and permanently hide valid rows behind them. Read the row ID before checking the blob pointer and add it to `droppedIds` when the blob is absent or empty. Please cover this with a corrupted-row regression test.

rahul-mixpanel (changes requested):
Requesting changes for four flush-path correctness and reliability issues. The bounded-batch approach is a good direction, but these cases can prematurely stop drainage, permanently stall a queue, or corrupt retry/backoff state. Please address them and add focused regression coverage before merging.

Sources/MPDB.swift
ketanmixpanel: I don't think that's an issue. The events remain persisted in the db and they will be sent to server in next flush cycle. 
Additionally I have added a check for the size of the payload when inserting event to DB. It will make sure to not add oversized records to db.

Sources/MixpanelInstance.swift
ketanmixpanel: fixed using max(id) approach.

Sources/FlushRequest.swift
ketanmixpanel: reverted the .timedOut based self.handleFlushFailure.

Sources/MPDB.swift
ketanmixpanel: fixed

MixpanelDemo/MixpanelDemoMacTests/MixpanelScreenTrackingTests.swift
```diff
@@ -28,9 +28,8 @@ class MixpanelScreenTrackingTests: MixpanelBaseTests {
...
```
rahul-mixpanel: why did we remove this from the tests?

MixpanelDemo/MixpanelDemoTVTests/TestConstants.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  TestConstants.swift
+//  MixpanelDemo
+//
+//  Created by Yarden Eitan on 6/28/16.
```
rahul-mixpanel: why is this old file showing up as a new file here?

Sources/Flush.swift
rahul-mixpanel: but the issue still exists if the flush is triggered before setting the server URL post init, right?

Sources/Flush.swift
```diff
@@ -9,7 +9,9 @@
 import Foundation
 
 protocol FlushDelegate: AnyObject {
-    func flush(performFullFlush: Bool, completion: (() -> Void)?)
+    /// Returns `true` if a flush was started, `false` if one was already in flight.
```
rahul-mixpanel: the comment is a bit confusing, can we have something like this?
> `true` if the flush started successfully; `false` if the request was rejected because another flush is already in progress.

rahul-mixpanel (changes requested):

MixpanelDemo/MixpanelDemoMacTests/MixpanelScreenTrackingTests.swift
ketanmixpanel: The screen tracking tests do not need to test the `screen_height`. These tests are frequently failing on CI due to screen_height is not available.

MixpanelDemo/MixpanelDemoTVTests/TestConstants.swift
ketanmixpanel: The file is copy of the `TestConstants.swift` from the iOS test target. Now all three targets have seperate TestConstants.swift file.

MixpanelDemo/MixpanelDemoMacTests/MixpanelScreenTrackingTests.swift
rahul-mixpanel: Okay. I think it’s worth checking why this happens if we’re seeing it occasionally. Screen height is one of the properties we track, so if it’s missing in some cases, we should verify whether there’s an underlying issue.

MixpanelDemo/MixpanelDemoTVTests/TestConstants.swift
rahul-mixpanel: Ok. The credits line above confused me. Would be good to remove or edit those.

Sources/Flush.swift
ketanmixpanel: fixed

Sources/Flush.swift
ketanmixpanel: We recommend in our docs now to use the MixpanelOptions for SDK initialization. `MixpanelOptions` has provision to set the serverUrl which will be used immediately from the SDK init. 

This fix is for the old init that we still support. If someone does not set serverUrl immediately after SDK init, and a flush happens by that time then events will go to default US endpoint.

MixpanelDemo/MixpanelDemoTVTests/TestConstants.swift
ketanmixpanel: fixed

Sources/MixpanelPersistence.swift
```diff
@@ -124,12 +135,9 @@ class MixpanelPersistence {
 
     func loadEntitiesInBatch(
         type: PersistenceType, batchSize: Int = Int.max, flag: Bool = false,
-        excludeAutomaticEvents: Bool = false
+        maxRowId: Int32? = nil
     ) -> [InternalProperties] {
-        var entities = mpdb.readRows(type, numRows: batchSize, flag: flag)
-        if excludeAutomaticEvents && type == .events {
-            entities = entities.filter { !($0["event"] as! String).hasPrefix("$ae_") }
-        }
```
rahul-mixpanel: why did we remove this filtering of automatic events?

rahul-mixpanel (dismissed):

Sources/MixpanelPersistence.swift
ketanmixpanel: I removed the filter due to below reasons -
1. As soon as we disable the auto tracked events, the SDK stops adding new auto tracked events to DB. 
2. As the events are already tracked, disabling the feature should not impact to it. The tracked events should go to the server.

Sources/MixpanelPersistence.swift
```diff
@@ -165,10 +171,11 @@ class MixpanelPersistence {
         mpdb.updateRowsFlag(.people, newFlag: !PersistenceConstant.unIdentifiedFlag)
     }
 
-    func resetEntities() {
-        for pType in PersistenceType.allCases {
-            mpdb.deleteRows(pType, isDeleteAll: true)
-        }
+    /// Deletes people updates made before any identify(). Everything else stays queued: events and
+    /// identified people rows already carry their distinct id, so they're sent under the user who
+    /// created them. Matches Android's reset, which clears only its anonymous-people table.
```
rahul-mixpanel: in general let us avoid quoting another platform and mentioning we're matching it. this is the ideal behaviour and all platforms should follow the same. I started doing the same on my PRs as well.

rahul-mixpanel (changes requested):

Sources/MixpanelPersistence.swift
ketanmixpanel: Makes sense.

